PR #207, opened 2019-01-08
Revit_Toolkit - Issue 206 - Default Materials

### LINKED ISSUE #206: Revit_Toolkit: Stop providing default materials
When pulling structural objects from Revit, the adapter currently provides a default material if nothing matching is found in the internal databases.

This is dangerous, as for an engineer using the toolkit probably will presume that everything has worked fine, even if in fact to match was found.

If no matching material is found, the toolkit should instead return an empty material with as many properties as can be found set, as well as a warning raised stating that a dummy material has been provided.

ZiolkowskiJakub:
Closes #206

### DIFF
--- a/Engine_Revit_UI/Compute/Warnings.cs
+++ b/Engine_Revit_UI/Compute/Warnings.cs
@@ -243,12 +243,36 @@ namespace BH.UI.Revit.Engine
 
         /***************************************************/
 
+        internal static void MaterialTypeNotFoundWarning(this FamilyInstance familyInstance)
+        {
+            string aMessage = "Matching BHoM ElementType could not be found.";
+
+            if (familyInstance != null)
+                aMessage = string.Format("{0} Element Id: {1}", aMessage, familyInstance.Id.IntegerValue);
+
+            BH.Engine.Reflection.Compute.RecordWarning(aMessage);
+        }
+
+        /***************************************************/
+
         internal static void InvalidDataMaterialWarning(this Material material)
         {
             string aMessage = "Material could not be correctly converted. Some BHoM Material data may not be valid.";
 
             if (material != null)
                 aMessage = string.Format("{0} Material Element Id: {1}", aMessage, material.Id.IntegerValue);
+
+            BH.Engine.Reflection.Compute.RecordWarning(aMessage);
+        }
+
+        /***************************************************/
+
+        internal static void InvalidDataMaterialWarning(this FamilyInstance familyInstance)
+        {
+            string aMessage = "Material could not be correctly converted. Some BHoM Material data may not be valid.";
+
+            if (familyInstance != null)
+                aMessage = string.Format("{0} Element Id: {1}", aMessage, familyInstance.Id.IntegerValue);
 
             BH.Engine.Reflection.Compute.RecordWarning(aMessage);
         }
--- a/Engine_Revit_UI/Compute/Warnings.cs
+++ b/Engine_Revit_UI/Compute/Warnings.cs
@@ -78,6 +78,18 @@ namespace BH.UI.Revit.Engine
 
         /***************************************************/
 
+        internal static void NullRevitElementWarning(this IBHoMObject BHoMObject)
+        {
+            string aMessage = "Referenced Revit element could not be find.";
+
+            if (BHoMObject != null)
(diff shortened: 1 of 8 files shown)


Engine_Revit_UI/Convert/Common/ToBHoM/Material.cs
```diff
@@ -98,5 +109,44 @@ internal static oM.Common.Materials.Material ToBHoMMaterial(this Material materi
...
+            StructuralAsset aStructuralAsset = aPropertySetElement.GetStructuralAsset();
+
+            if (aStructuralAsset.Behavior != StructuralBehavior.Isotropic)
+                return;
+
+            material_Destination.YoungsModulus = UnitUtils.ConvertFromInternalUnits(aStructuralAsset.YoungModulus.X, DisplayUnitType.DUT_PASCALS);
+            material_Destination.Density = UnitUtils.ConvertFromInternalUnits(aStructuralAsset.Density, DisplayUnitType.DUT_KILOGRAMS_PER_CUBIC_METER);
+            material_Destination.CoeffThermalExpansion = UnitUtils.ConvertFromInternalUnits(aStructuralAsset.ThermalExpansionCoefficient.X,DisplayUnitType.DUT_INV_CELSIUS);
+            material_Destination.PoissonsRatio = aStructuralAsset.PoissonRatio.X;
+            material_Destination.CompressiveYieldStrength = UnitUtils.ConvertFromInternalUnits(aStructuralAsset.ConcreteCompression, DisplayUnitType.DUT_PASCALS);
```
IsakNaslundBh: Would assume this property only exist for concrete materials. Would be good to add a check what material type it is, and add the appropriate design parameters, at least for steel and concrete.

Engine_Revit_UI/Convert/Common/ToBHoM/Material.cs
```diff
@@ -98,5 +109,44 @@ internal static oM.Common.Materials.Material ToBHoMMaterial(this Material materi
...
+            if (aElementId == null || aElementId == ElementId.InvalidElementId)
+                return;
+
+            Document aDocument = material_Source.Document;
+
+            PropertySetElement aPropertySetElement = aDocument.GetElement(aElementId) as PropertySetElement;
+
+            StructuralAsset aStructuralAsset = aPropertySetElement.GetStructuralAsset();
+
+            if (aStructuralAsset.Behavior != StructuralBehavior.Isotropic)
```
IsakNaslundBh: Could we add a warning if this happens, to alert users that they are getting a zeroed out material?

Engine_Revit_UI/Convert/Common/ToBHoM/Material.cs
```diff
@@ -98,5 +109,44 @@ internal static oM.Common.Materials.Material ToBHoMMaterial(this Material materi
...
         /***************************************************/
+        /****             Private methods               ****/
+        /***************************************************/
+
+        private static void Update(this oM.Common.Materials.Material material_Destination, Material material_Source)
+        {
+            if (material_Source == null)
+                return;
+
+            ElementId aElementId = material_Source.StructuralAssetId;
```
IsakNaslundBh: Could we add a warning if this happens, to alert users that they are getting a zeroed out material?

Engine_Revit_UI/Convert/Common/ToBHoM/Material.cs
ZiolkowskiJakub: If Materials are common for all disciplines then we need to agree which parameters will be filled in... I strongly believe the data such as Density should be copied for all materials.

IsakNaslundBh: Sorry, was not clear enough! Only meant the very last row!

Engine_Revit_UI/Compute/Warnings.cs
```diff
@@ -78,6 +78,18 @@ internal static void NullObjectWarning()
 
         /***************************************************/
 
+        internal static void NullRevitElementWarning(this IBHoMObject BHoMObject)
+        {
+            string aMessage = "Referenced Revit element could not be find.";
```
IsakNaslundBh:
```suggestion
            string aMessage = "Referenced Revit element could not be found.";
```

IsakNaslundBh:
All working well on my end now, thanks for fixing!

PR #246, opened 2019-02-22
Revit_Toolkit - Issue 244 - Include Non Visible Objects

### LINKED ISSUE #244: Revit_Toolkit: Option to pull invisible lines
For some element types, invisible lines will need to be pulled. Option to toggle between pulling or not would be good to have.

ZiolkowskiJakub:
Include Non Visible Objects when pulling edges of Revit Element

Closes #244

### DIFF
--- a/Revit_Engine/Modify/SetPullEdges.cs
+++ b/Revit_Engine/Modify/SetPullEdges.cs
@@ -51,6 +51,16 @@ namespace BH.Engine.Adapters.Revit
             return aFilterQuery;
         }
 
+        [Deprecated("2.2")]
+        [Description("Sets Pull Edges option for FilterQuery.")]
+        [Input("filterQuery", "FilterQuery")]
+        [Input("pullEdges", "Set to true if include geometry edges of Revit Element in CustomData of BHoMObject")]
+        [Output("FilterQuery")]
+        public static FilterQuery SetPullEdges(this FilterQuery filterQuery, bool pullEdges)
+        {
+            return SetPullEdges(filterQuery, pullEdges, false);
+        }
+
         /***************************************************/
     }
 }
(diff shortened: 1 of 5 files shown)


Revit_Engine/Modify/SetPullEdges.cs
```diff
@@ -36,18 +36,17 @@ public static partial class Modify
         [Description("Sets Pull Edges option for FilterQuery.")]
         [Input("filterQuery", "FilterQuery")]
         [Input("pullEdges", "Set to true if include geometry edges of Revit Element in CustomData of BHoMObject")]
+        [Input("includeNonVisibleObjects", "Set to true if include non visible objects of Revit Element in geometry")]
         [Output("FilterQuery")]
-        public static FilterQuery SetPullEdges(this FilterQuery filterQuery, bool pullEdges)
+        public static FilterQuery SetPullEdges(this FilterQuery filterQuery, bool pullEdges = true, bool includeNonVisibleObjects = false)
```
IsakNaslundBh: One comment here regarding backwards compatibility for old scripts.

When modifying a public method like this, by adding inputs, please keep the old method as well, and make it call this one with the default attributes. Also put a `Deprecated` tag on it. This is to avoid breaking existing scripts making use of this method. [see here](https://github.com/BHoM/documentation/wiki/Backwards-compatibility#methodscomponents)

This would be adding 

```
        [Deprecated("2.2")]
        public static FilterQuery SetPullEdges(this FilterQuery filterQuery, bool pullEdges = true)
        {
            return SetPullEdges(filterQuery, pullEdges, false);
        }
```

To what is already in the file.

IsakNaslundBh:
Looks good, thanks Jakub!

PR #268, opened 2019-03-14
Revit_Toolkit - Issue 266 - Physical Objects

### LINKED ISSUE #247: Revit_Toolkit: Update Push of Environment Element using new extended properties approach
As we discussed during Revit Toolkit Planning implementation of extended properties disabled current push capability for all elements.
This need to be re-implemented before completing additional push elements that are logged as separate Issues already.

### LINKED ISSUE #265: Revit_Toolkit: Column Framing elements can't be pushed to Revit
Currently it doesn't seem possible to push columns, digging into the code I think it is due to the familyplacement type. I think a column should be able to be placed/defined as vertical or slanted which is why I think it is failing.

image

Error from push component:
1. BHoM Object location does not match with the required placement type of Revit family BHoM Guid: 8fbb9612-e44a-4bb4-bf5b-d7804065cd1e Element Id : 630797

### LINKED ISSUE #266: Revit_Toolkit: Physical Objects Implementation

### LINKED ISSUE #270: Revit_Toolkit: Rename Revit Specific objects
DraftingObject -> DraftingInstance (represenation of generic annotations, 2D objects etc. )
GenericObject -> ModelInstance (represenation of 3D model objects)
ObjectProperties -> InstanceProperties (equivalent of ElementType in Revit API)

### LINKED ISSUE #271: Revit_Toolkit: Family Object implemenation
Revit_Toolkit specific object.

Class to store Family specific data and allow user to push/pull objects.

ZiolkowskiJakub:
Pull for Architectural Wall, Roof, Floor implemented
Push for Architectural Wall, Roof, Floor implemented
Warnings for Analytical objects (Environmental) added

Closes #266 
Closes #247 
Closes #270 
Closes #271 
Closes #265

### Changes
* Implementation of Physical Objects (Wall, Floor, Roof)
* Update Push of Environment Elements
* Implementation of Family Object (Class to store Family specific data and allow user to push/pull objects) 
* Revit_Toolkit object rename (DraftingObject -> DraftingInstance, GenericObject -> ModelInstance, ObjectProperties -> InstanceProperties)
* Push bug fix for Structural Columns/Framing


michaldengusiak:
1. Pull Wall Env - OK


2. Pull Wall Arch- OK


3. Roof - Env - OK


4. Roof - Arch- Fail


5. Pull Wall Arch1- OK


5. Pull Wall Env1- OK


6. Pull Floor Env- OK


6. Pull Floor Arch- OK


please let me know when free so we can set up push as I can not get it working  right now

Revit_Engine/Create/Elements/Floor.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates Floor object.")]
+        [Input("object2DProperties", "Object2DProperties")]
+        [Input("edges", "External edges of Floor")]
+        [Output("Floor")]
+        public static Floor Floor(Object2DProperties object2DProperties, ICurve edges)
```
IsakNaslundBh: These kind of constructors should not be part of the Revit_Engine. This is generic and should sit in the Architecture_Engine.

Only very Revit specific creates should be part of the create section here.

https://github.com/BHoM/BHoM_Engine/issues/841

Revit_Engine/Create/Elements/Roof.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates Roof object.")]
+        [Input("object2DProperties", "Object2DProperties")]
+        [Input("edges", "External edges of Roof")]
+        [Output("Roof")]
+        public static Roof Roof(Object2DProperties object2DProperties, ICurve edges)
```
IsakNaslundBh: Same as for floor. No methods like this in Revit_Engine

Revit_Engine/Create/Elements/Wall.cs
```diff
@@ -0,0 +1,97 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates Wall object.")]
+        [Input("object2DProperties", "Object2DProperties")]
+        [Input("locationCurve", "Location curve of the wall (bottom edge)")]
+        [Input("height", "Wall height")]
+        [Output("Wall")]
+        public static Wall Wall(Object2DProperties object2DProperties, ICurve locationCurve, double height)
```
IsakNaslundBh: As above

Revit_oM/Elements/DraftingInstance.cs
```diff
@@ -20,19 +20,20 @@
...
 
+using BH.oM.Adapters.Revit.Interface;
 using BH.oM.Adapters.Revit.Properties;
 using BH.oM.Base;
 using BH.oM.Geometry;
 
 namespace BH.oM.Adapters.Revit.Elements
 {
-    public class DraftingObject : BHoMObject
+    public class DraftingInstance : BHoMObject, IInstance
```
IsakNaslundBh: Any specific reason for this change of name? Not saying that it is wrong, but might break scripts out in production.

Revit_oM/Elements/DraftingInstance.cs
```diff
@@ -20,19 +20,20 @@
...
 {
-    public class DraftingObject : BHoMObject
+    public class DraftingInstance : BHoMObject, IInstance 
     {
         /***************************************************/
         /**** Public Properties                        ****/
         /***************************************************/
 
-        public ObjectProperties ObjectProperties = new ObjectProperties();
+        public InstanceProperties InstanceProperties { get; set; } = new InstanceProperties();
```
IsakNaslundBh: Agreeing on changing the name for this away from the very generic `ObejctProperties`, but `InstanceProperties` is almost as generic. Would something like `DraftingProperties` not be clearer for the name of the class? Might be wrong here though if this is used for multiple things

Also I would call the name of the property just properties, as you already know what kind of object you are using. Something like:

```
public DraftingProperties Properties { get; set; } = new DraftingProperties ();
```

I think would make more sense

Revit_Engine/Create/Elements/Floor.cs
ZiolkowskiJakub: I fully agree on that however can we first fully agree on IObject2D, Wall etc. structure (methods, properties etc.) before we spread the changes across to different repositories? I assume this is proof of concept for now.

Revit_Engine/Create/Elements/Roof.cs
ZiolkowskiJakub: I fully agree on that however can we first fully agree on IObject2D, Wall etc. structure (methods, properties etc.) before we spread the changes across to different repositories? I assume this is proof of concept for now.

Revit_Engine/Create/Elements/Wall.cs
ZiolkowskiJakub: As above

Revit_oM/Elements/DraftingInstance.cs
ZiolkowskiJakub: Yes... I agree it may influence existing scripts however I strongly believe that currently it is really rare to use these objects directly (calling it via type name). The reason for this change is that I was trying to find commonality between Revit and BHoM and hopefully "Instance" name is more understandable than  "Object" to Revit users

Revit_oM/Elements/DraftingInstance.cs
ZiolkowskiJakub: Correct.

ZiolkowskiJakub: Just to add on that. I always assume inputs should be named in such way user do not have to think to much what should be connected to it... so if input is called "InstanceProperties" you know straight away what kind of input you need to connect... if you start searching for just "Properties" it may appear there is many objects with such a name.

michaldengusiak:
When I have rooflight it take roflight and not roof geometry..

al-fisher:
@michaldengusiak @IsakNaslundBh can you re-look at this to approve? Thanks

michaldengusiak:
@IsakNaslundBh  please have a look and once you happy approve and then I will run final test

jdarcyBHE:
Bug fix for column Push works for both Revit 2018 and 2019

michaldengusiak:
unfortunately some fix still needed...

1. Pull Wall Env - OK for Surface, Fail for Edges


2. Pull Wall Arch- Fail


3. Roof - Env - OK for Surface, Fail for Edges


4. Roof - Arch- Fail


5. Pull Wall Arch1- Fail


5. Pull Wall Env1- OK for Surface, Fail for Edges


6. Pull Floor Env- OK for Surface, Fail for Edges


6. Pull Floor Arch- Fail


I got feeling that some of them were working fine before I am on:
`Revit_Toolkit-Issue226-PhysicalObjects`
`Dynamo_Toolkit-Issue133-PlanarSurfaceConversion`
I rebuild all with CodeAsisstant now with try with FullBHoM as can not believe in above

IsakNaslundBh:
To comment on this:

If the architectural elements are failing here, it is sort of ok, as they will be deleted anyway and replaced with the new physical objects. We will need to get this merged/closed quite soon to make way for changes in env oM and structures oM.

PR #299, opened 2019-05-23
Revit_Toolkit-Issue284-PullSpaceTypes

### LINKED ISSUE #284: Revit_Toolkit:  Pull Space Types
When you pull a Space, its parameter for "Space Type" returns just the ID of the Space Type.  When you want to pull a "Space Type Settings" category (which is the category name for "Space Types"), nothing is returned.

ZiolkowskiJakub:
Fix for CategoryFilter and TypeNameFilter. Enable to pull Space Types and similar kinds of objects

Closes #284 

### Test files
[Dynamo script and Sample Model](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Revit_Toolkit/Revit_Toolkit-Issue284-PullSpaceTypes?csf=1&e=O6c3fg)

### DIFF
--- a/Engine_Revit_UI/Query/BuiltInCategory.cs
+++ b/Engine_Revit_UI/Query/BuiltInCategory.cs
@@ -156,6 +156,12 @@ namespace BH.UI.Revit.Engine
                 if (aCategory.Name == categoryName)
                     return (BuiltInCategory)aCategory.Id.IntegerValue;
 
+            switch(categoryName)
+            {
+                case "Space Type Settings":
+                    return Autodesk.Revit.DB.BuiltInCategory.OST_HVAC_Load_Space_Types;
+            }
+
             return Autodesk.Revit.DB.BuiltInCategory.INVALID;
         }
 
--- a/Engine_Revit_UI/Query/ElementIds.cs
+++ b/Engine_Revit_UI/Query/ElementIds.cs
@@ -271,7 +271,14 @@ namespace BH.UI.Revit.Engine
             }
 
             if (aType == null)
-                return null;
+            {
+                foreach(TypeInfo aTypeInfo in aAssembly.DefinedTypes)
+                    if(aTypeInfo.Name == typeName)
+                    {
+                        aType = aTypeInfo.AsType();
+                        break;
+                    }
+            }
 
 
             return new FilteredElementCollector(document).OfClass(aType).ToElementIds();


ZiolkowskiJakub:
You are able to pull Spece types in two ways:
- Using TypeNameFilterQuery
- Using CategoryFilterQuery 

image


image

FraserGreenroyd:
Adding additional reviewers from both Environments and Structures to aid this PR :smile:

michaldengusiak:
however Space types are not pulled for spaces...
Should I log as separate issue?
image

ZiolkowskiJakub:
Hi @michaldengusiak,

This PR allows user to pull Space Type from document. It also takes default approach of Revit_Toolkit that Parameters values are stored in CustomData of BHoMObject. Therefore Space Type  is stored in CustomData:

image

This PR does not implement any changes to Environmental or Structure objects as it was not reported to do so.

Please open new Issue if we need to update Environmental objects.

FraserGreenroyd:
@ZiolkowskiJakub not sure how that follows - `Space` is an Environmental Object so how did this PR not implement changes to `Environmental` or `Structural` objects?

ZiolkowskiJakub:
Hi @FraserGreenroyd 

It does not due to fact the data stored as ElementId was there before this PR. The only issue was that you could not pull Revit object from document using ElementId stored in Space CustomData.

There was no change to Environmental Space object because related data is stored in CustomData and there is generic approach to populate CustomData of BHoMObject.

FraserGreenroyd:
@ZiolkowskiJakub ok that makes a bit of sense, though if the object being returned is the `Environment Space` I would argue that we should be populating the `Space` properties either in paramount to or in addition with the `CustomData`.

Also, as an FYI, it might be worthwhile having a chat with @IsakNaslundBh and I regarding CustomData - if you look at the planning boards around Fragments [here](https://github.com/BHoM/BHoM/projects/2) you'll note phase 3a and 3b regarding the removal of `CustomData`. Obviously Revit Toolkit is a strong user of `CustomData` so it would be good to make sure you're aware of those plans and able to feedback your thoughts during that development process :smile:

ZiolkowskiJakub:
Hi @FraserGreenroyd,

Thank you very much for the information. 

 I assume that te change will looks something like this for BHoMObject:

This:
```
public Dictionary<string, object> CustomData { get; set; } = new Dictionary<string, object>();
```
will be replaced by this:
```
public List<IBHoMFragment> FragmentProperties { get; set; } = new List<IBHoMFragment>();
```
Am I right?

PR #417, opened 2019-10-19
Updating default configurations for debug and release

### LINKED ISSUE #416: Ensure default (debug and release) configurations build
Following the recent close out of PR switching dependency on NuGets [here](https://github.com/BHoM/Revit_Toolkit/pull/414) default `debug` and `release` no longer compile.

Suggested fix is to simply align the default `debug` and `release` to one selected versioned config.

This is causing auto-installer builds on azure to fail. 
There is asymmetry between AppVeyor which only checks the specific hard coded configs on each repo and Azure which is in addition also building the available default `release`

Will raise issues to resolve differences in CI approach - however, think important that any default config must be buildable in any case.

al-fisher:

Closes #416 

Following update of NuGets - have aligned the default `debug` and `release` to `debug2018` and `release2018` configurations to ensure any default set ups can also build.


al-fisher:
Fantastic - thanks both @FraserGreenroyd @IsakNaslundBh

PR #424, opened 2019-10-29
Fix Pull Space

### LINKED ISSUE #423: Issue with pulling Spaces
#### Description:

Environmental Engine method `aSpace.AddFragment(aOriginContextFragment)` has been chagned to not accept similar types 

image

michaldengusiak:

Fixes #423 

### Test files

[test file](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Revit_Toolkit/Revit_Toolkit-Issue423-PullSpaces?csf=1&e=4n2MgL)

### Changelog

* fix fragments to pull space

### DIFF
--- a/Engine_Revit_UI/Convert/Environment/ToBHoM/Space.cs
+++ b/Engine_Revit_UI/Convert/Environment/ToBHoM/Space.cs
@@ -90,7 +90,7 @@ namespace BH.UI.Revit.Engine
 
             SpaceAnalyticalFragment aSpaceAnalyticalFragment = new SpaceAnalyticalFragment();
             aSpaceAnalyticalFragment = aSpaceAnalyticalFragment.UpdateValues(pullSettings, spatialElement) as SpaceAnalyticalFragment;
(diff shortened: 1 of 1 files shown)


ZiolkowskiJakub:
Hi @FraserGreenroyd,

> Taking the issue at face value, it would suggest to me there is a problem with the core Environment_Engine (or, more likely now, Base_Engine) adding fragments than a problem with Revit using them.

I am sure the issue is related only to Revit_Toolkit. The limitation added in Environment_Engine cause that the same Fragments cannot be added and because of that it rise error in Revit_Toolkit before element has been translated to BHoM. I strongly believe adding the same Fragments twice (or more) shall not be allowed so updated behaviour of Enviroment_Engine is appropriate. Nevertheless, it needs some adjustments in Revit_Toolkit.

IsakNaslundBh:
Should not be affecting any of the structural objects, so I am fine with this if you guys are.

PR #464, opened 2019-11-19
Batch cleanup and fix for #440

### LINKED ISSUE #440: Revit_Toolkit: Make all BH.UI.Revit.Engine internal methods public
#### Description:
It would be good to change all public methods in `BH.UI.Revit.Engine` to internal so that there is one pattern across the whole project.

pawelbaran:

Closes #440

### Test files
Not needed.

### Changelog
- all `internal` methods in `BH.UI.Revit.Engine` that are possibly useful outside Revit_Toolkit were made `public`

### Additional comments
I made another iteration of the batch cleanup - looking forward to getting a few of these in the near future.

### DIFF
--- a/Adapter_Revit_UI/AdapterActions/Push.cs
+++ b/Adapter_Revit_UI/AdapterActions/Push.cs
@@ -33,7 +33,7 @@ namespace BH.UI.Revit.Adapter
     {
 
         /***************************************************/
-        /**** Protected Methods                         ****/
+        /****             Protected Methods             ****/
         /***************************************************/
 
         public override List<IObject> Push(IEnumerable<IObject> objects, string tag = "", Dictionary<string, object> config = null)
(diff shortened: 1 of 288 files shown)


pawelbaran:
> Note that the convert methods don't seem to be in the right namespace (and hence don't show up in the UI). I think should be `BH.Engine.Adapters.Revit.Query`. With making all these methods public we are naturally exposing a lot into the UI but a) the converts aren't visible anyway at the moment b) I don't think they're widely used in UI's and c) we can always suppress these in some UI's as and where necessary/appropriate.

All convert/modify/query methods that reference Revit API are stored in `BH.UI.Revit.Engine` on purpose: Revit API cannot be used outside the Revit context, so there is no benefit of exposing them in the UI (they will all spit errors). However, changing the access level from `internal` to `public` makes them available for other projects, which might be useful in some cases.

PR #551, opened 2020-02-20
Boilerplate ToBHoMConstruction removed

pawelbaran:
### Issues addressed by this PR

Closes #542

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue542%2DUnnecessaryToBHoMConstruction) - this will work with any (even empty) Revit file.

### Changelog

- `ToBHoMConstruction` method has been unified and made work on any types that inherit from `Autodesk.Revit.DB.HostObjAttributes` when converted to BHoM

PR #552, opened 2020-02-20
Change method name for compliance

### LINKED ISSUE #537: Revit_Toolkit: ToBHoMObjects(ProjectInfo) method is incompliant
#### Broken rules:

- File name is unrelated to method name
- There are multiple hard-coded parameter names that should be migrated to `MapSettings`

FraserGreenroyd:
Fixes #537 

### Test files
Existing and code review

### DIFF
--- a/Engine_Revit_UI/Convert/Environment/ToBHoM/Building.cs
+++ b/Engine_Revit_UI/Convert/Environment/ToBHoM/Building.cs
@@ -104,11 +104,7 @@ namespace BH.UI.Revit.Engine
 
                 //Set identifiers & custom data
                 building = building.SetIdentifiers(document.ProjectInformation) as Building;
-                building = building.SetCustomData("Time Zone", timeZone) as Building;
-                building = building.SetCustomData("Place Name", placeName) as Building;
-                building = building.SetCustomData("Weather Station Name", weatherStationName) as Building;
 
-                building = building.SetCustomData("Project Angle", projectAngle) as Building;
                 building = building.SetCustomData("Project East/West Offset", projectEastWestOffset) as Building;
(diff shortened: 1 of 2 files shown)


FraserGreenroyd:
> `FromRevit` is even less compliant than `ToBHoMObjects` imo... All Revit->BHoM convert names start with _ToBHoM_ prefix, my point in the issue was that `ToBHoMObjects` does not tell what sort of convert is that (it could be literally anything), while the filename is quite explicit. Maybe it could be something like `ToBHoMBuilding`?

`ToBHoM` is incompliant to the guidelines, all convert methods should start with `To` or `From` and not be `ToBHoM` or `FromBHoM` because BHoM is implicit, so the conversion should always be `ToSoftware` or `FromSoftware`.

> BTW, this will conflict with #551, which was raised a few minutes earlier 🐒

Whoever gets approvals first wins :wink:

pawelbaran:
> > BTW, this will conflict with #551, which was raised a few minutes earlier 🐒
> 
> Whoever gets approvals first wins 😉

I will block this one forever then ⚔️ 

And seriously, is that stated somewhere that the names should be `ToSoftware` or `FromSoftware`? If it is a general, stated rule, then I am happy to align - in such case I would opt for closing this PR and renaming all methods in one, isolated action.

pawelbaran:
Thanks @FraserGreenroyd 
[Convert Method Naming Convention](https://github.com/BuroHappoldEngineering/BuroHappold_documentation/wiki/IsValidConvertMethodName)

FraserGreenroyd:
Discussed with @pawelbaran over Skype and shown him compliance documentation that fulfils the above.

Am happy to go with a mass renaming, so have reverted that change in this PR, and will have a separate issue raised for compliance of Convert method names.

PR #555, opened 2020-02-20
FilterByType method deprecated

### LINKED ISSUE #543: Revit_Toolkit: BH.Engine.Adapters.Revit.Query.FilterByType is never used
#### Description:
As in the title - the method could be deleted.

pawelbaran:

Closes #543 

### Test files

Not needed

### Changelog

- `BH.Engine.Adapters.Revit.Query.FilterByType(IEnumerable<object>, Type)` has been deprecated as it is a duplicate of same method in `BH.Engine.Base.Query`

### DIFF
--- a/Revit_Engine/Query/FilterByType.cs
+++ b/Revit_Engine/Query/FilterByType.cs
@@ -38,6 +38,7 @@ namespace BH.Engine.Adapters.Revit
         /****              Public methods               ****/
         /***************************************************/
 
+        [Deprecated("3.1", "This is a duplicate of BH.Engine.Base.Query.FilterByType method.", typeof(BH.Engine.Base.Query), "FilterByType")]
         [Description("Filters objects by type.")]
         [Input("objects", "Objects to be filtered by Type")]
         [Input("type", "Type")]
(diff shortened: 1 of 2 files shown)


Revit_Engine/Query/FilterByType.cs
```diff
@@ -38,6 +38,7 @@ public static partial class Query
         /****              Public methods               ****/
         /***************************************************/
 
+        [Deprecated("3.1", "This is a duplicate of BH.Engine.Base.Query.FilterByType method.", typeof(BH.Engine.Base.Query), "FilterByType")]
```
FraserGreenroyd: Was anywhere in the Revit TK using this method? Does it need converting over to the Base Engine query? Cause I can only see two files changed here?

pawelbaran: I made an effort to deprecate instead of deleting because this method has been exposed to the UI (it is Revit_Engine). If not that, I would be more than happy to delete it straight away.

FraserGreenroyd: Ok cool if nothing is using it code wise then that's fine, that's all I wanted to check :smile:

PR #569, opened 2020-02-26
BH.Engine.Adapters.Revit.Query.IsZero(double) deprecated

### LINKED ISSUE #532: Revit_Toolkit: Deprecate BH.Engine.Adapters.Revit.Query.IsZero(double)
#### Description:
The approach with hard-coded tolerance in the method is not correct. This should be done differently (going up the reference chain, it is used to find top faces of solids - changes should probably be introduced on that level already).

pawelbaran:

Closes #532

### Test files

Not needed

### Changelog
- `BH.Engine.Adapters.Revit.Query.IsZero(double)` method has been deprecated

### DIFF
--- a/Revit_Engine/Query/IsZero.cs
+++ b/Revit_Engine/Query/IsZero.cs
@@ -33,7 +33,7 @@ namespace BH.Engine.Adapters.Revit
         /****              Public methods               ****/
         /***************************************************/
         
-        [DeprecatedAttribute("3.1", "This method is questionably useful and potentially harmful as it introduces a hardcoded tolerance.")]
+        [DeprecatedAttribute("3.1", "This method should not be used as it introduces a hardcoded tolerance.")]
         [Description("Checks if given double value is almost equal 0 (MicroDistance Tolerance).")]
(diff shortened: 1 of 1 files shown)


Revit_Engine/Query/IsZero.cs
```diff
@@ -32,7 +32,8 @@ public static partial class Query
         /***************************************************/
         /****              Public methods               ****/
         /***************************************************/
-
+        
+        [DeprecatedAttribute("3.1", "This method is questionably useful and potentially harmful as it introduces a hardcoded tolerance.")]
```
FraserGreenroyd: As much as I find the description funny, could we perhaps go for something slightly less abrasive to the user? :smile:

pawelbaran: I've heard honesty is a virtue 🐒 I tried to make it more politically correct yet informative - do you like it more after the change?

pawelbaran:
After discussing with @FraserGreenroyd, we concluded it is safe to merge regardless the appveyor fail (it seems not to be related with the code changes).

PR #577, opened 2020-03-05
UpdateOnly PushType unlocked

### LINKED ISSUE #460: Revit_Toolkit: Revise Move methods
#### Description:
There is a few `Move` methods in the Revit_Toolkit project:
- they should possibly be stored in one place
- they should be named `Translate`
- other transformations should be added too
- the line below should possibly be replaced with a transform incl. rotation too
- hardcoded `Tolerance.Micro` should be replaced with `RevitSettings.DistanceTolerance`
https://github.com/BHoM/Revit_Toolkit/blob/a0e5c4f4c747d8cd59eb63ffc21f0407eab39db8/Adapter_Revit_UI/CRUD/Create.cs#L213

As a follow up, `BH.UI.Revit.Engine.Query.IsSimilar` method should be squashed.

### LINKED ISSUE #546: Revit_Toolkit: Prototype of UpdateOnly PushType
#### Description:
Following #545 and #522, it would be great to introduce a prototype of Update push that does not delete the existing item, but modifies it instead.

### LINKED ISSUE #575: Revit_Toolkit: Delete FamilyLoadOptions or move it to a separate file
#### Description:

As in the title - the class needs to be either deleted or moved to a separate file.

### LINKED ISSUE #576: Revit_Toolkit: Enable setting enum parameters based on string
#### Description:

Most parameters in Revit have open formula (any value can be set to them), but a limited group is closed (the user can choose from predefined values), example in the screenshot below.

image

The information about values of the latter is stored as an `enum` in a way, where the UI shows a string, while programatically we operate on their correspondent `ints`. Unfortunately, regardless >10 years (?) or requests on fora, Autodesk does not feel like sharing the arcane knowledge of which `enum` in the API is correspondent to which parameter, therefore setting a value means a guessing game, where one needs to pop ints until the right value is found.

Still, it would be great to enable setting the enums based on strings since otherwise one will not be able to e.g. set beam alignment from script level without guessing int after int (Dynamo style).

pawelbaran:

Closes #460
Closes #546 
Closes #575 
Closes #576

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue546%2DUpdateOnly)

### Changelog

- `UpdateOnly` `PushType ` has been unlocked
- Basic `Update` methods have been added
- `Move` methods were refactored and renamed to `SetLocation`
- Enum parameter queries were changed to return string representation of the enum

### Additional comments

This PR provides a framework for `Update` with a few prototype methods. These may contain bugs, particularly when it comes to setting location - mutual relationship between the parameters and geometrical properties of objects in Revit is an eternal pain that I am aiming to sort out once and for all in the next milestone. I would be happy to merge this PR and raise issues if any bugs are found instead of blocking.

Concerning #576, the implemented solution is a brute force trial/error solution that causes a massive slowdown (~10ms per element per parameter) - unfortunately, it seems to be the only way to go. I am comfortable with it because the slowdown appears only when the user attempts to change the enum parameter based on string, which otherwise would not work at all.


Engine_Revit_UI/Modify/Update.cs
```diff
@@ -0,0 +1,124 @@
...
+using BH.oM.Base;
+using System;
+
+namespace BH.UI.Revit.Engine
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /****              Private Methods              ****/
+        /***************************************************/
```
alelom: check this

Adapter_Revit_UI/AdapterActions/Push.cs
```diff
@@ -109,11 +108,16 @@ public override List<object> Push(IEnumerable<object> objects, string tag = "",
...
+                        Element element = obj.Element(document);
+                        if (element != null)
+                            success |= Update(element, obj, RevitSettings);
+                    }
                 }
                 else if (pushType == PushType.FullCRUD)
                 {
-                    BH.Engine.Reflection.Compute.RecordError("Full CRUD is currently not supported by Revit_Toolkit, please use DeleteThenCreate instead.");
+                    BH.Engine.Reflection.Compute.RecordError("Full CRUD is currently not supported by Revit_Toolkit, please use Create, UpdateOnly or DeleteThenCreate instead.");
                 }
```
alelom: This could be moved up around line 63, because it's a blocking check.

Adapter_Revit_UI/AdapterActions/Push.cs
```diff
@@ -73,7 +72,7 @@ public override List<object> Push(IEnumerable<object> objects, string tag = "",
             if (UIControlledApplication != null && pushConfig.SuppressFailureMessages)
                 UIControlledApplication.ControlledApplication.FailuresProcessing += ControlledApplication_FailuresProcessing;
 
-            
+
             // If unset, set the pushType to AdapterSettings' value (base AdapterSettings default is FullCRUD).
             if (pushType == PushType.AdapterDefault)
                 pushType = PushType.DeleteThenCreate;
```
alelom: If you do https://github.com/BHoM/Revit_Toolkit/pull/577/files#r388440231,
then this could be put just after that, so we have an initial setup section that checks/sets the PushType.

Engine_Revit_UI/Modify/Update.cs
alelom: Nevermind, I now got why it's like this.

Adapter_Revit_UI/AdapterActions/Push.cs
pawelbaran: I moved the blocking check as far up as possible, should be fine now

kThorsager:
### Grid:
success = true on a fail?
image

### Level:
Error when updating the elevation: 
`1. Unable to set location of Revit element of type Autodesk.Revit.DB.Level based on BHoM object of type BH.oM.Geometry.SettingOut.Level beacuse no suitable method could be found. Revit ElementId: 715860 BHoM_Guid: acbc186a-498a-4c75-997d-121ca7509958` 
despite working fine

The same error message persists when updating the name, which works.

### Standard Column:
Update double parameter: no error and success = true when failing
image

Preforming thease after one another causes unwanted results.
image
image
Where the first twist seem to persist when updatning the column to vertical. 
image
The Column which is returned by the Push component does not reflect this.
This is a recurring thing, that the returned `objects` show the input `objects` with the change applied
and the object which is updated in Revit is a mishmash of various updates.

Update top level for column:
No indication that this did not work for the user:
image

Material:
No idea how to verify this:
image

Update Enum:
Does not work, potentially due to stacking to many updates.
image
image

Update bool parameter:
Can neither find the parameter before nor after:
image

### Standard Beam:
These updates function like one would expect, (not like the columns) where the update is the whole element. i.e. The object in Revit is updated to the object one provides to the push component.

### Pinned Element:
Fails due to being pinned, nice.
returns success = true ?
image

### Update based on Analytical:
returns success = true on fail
image


### General Testing
Does not appear to break other parts of the toolkit.


You know better than I if these are expected behaviour, future issues, or to be handled here @pawelbaran

pawelbaran:
Thanks for the review @kthorsager, the two big things seem to be:
- update of location-related parameters - the way in which Revit manages that is so overcomplicated that it is hard to even record an error without writing some 100 lines of checks and checks of checks. As mentioned in the PR description, that is something that I am planning to challenge in the next quarter (I hope I will come up with a sensible way of handling that problem), therefore I decided not to put too much effort now.
- true/false on output - to be honest, it does not make sense to me to return a single Boolean when updating a list of elements:
    - What if one element gets updated correctly and the other does not?
    - What if 90 parameters get updated correctly and one not?

    On top of that, the adapter seems not to allow manipulating the input collection, which means that when something goes wrong, I cannot update the output objects to reflect the changes that have/have not been made. @alelom please correct me if I am confabulating.

    All in all, it brought me to the point where I just started returning anything (mostly `true`). This sounds like a longer discussion that may touch some of the core BHoM repos, therefore I am leaning towards letting it go in this PR + raising a follow-up issue.

Besides that, a few comments below:

> ### Level:
> Error when updating the elevation:
> `1. Unable to set location of Revit element of type Autodesk.Revit.DB.Level based on BHoM object of type BH.oM.Geometry.SettingOut.Level beacuse no suitable method could be found. Revit ElementId: 715860 BHoM_Guid: acbc186a-498a-4c75-997d-121ca7509958`
> despite working fine
> 
> The same error message persists when updating the name, which works.

Good catch, fixed.

> 
> ### Standard Column:
> Preforming thease after one another causes unwanted results.
> image
> image
> Where the first twist seem to persist when updatning the column to vertical.
> image
> The Column which is returned by the Push component does not reflect this.
> This is a recurring thing, that the returned `objects` show the input `objects` with the change applied
> and the object which is updated in Revit is a mishmash of various updates.

Fixed, should be better (fine?) now.

> 
> Material:
> No idea how to verify this:
> image

Sorry, I have updated the test file to make it more informative.

> 
> Update Enum:
> Does not work, potentially due to stacking to many updates.
> image
> image
 
This seems to be me fixing one thing and breaking another - should be fixed now.

> Update bool parameter:
> Can neither find the parameter before nor after:
> image

This guys sits here:
image

If you cannot find it, you could also try changing some other, randomly chosen checkbox (which might be even better as creating a new test case).

IsakNaslundBh:
> > * On top of that, the adapter seems not to allow manipulating the input collection, which means that when something goes wrong, I cannot update the output objects to reflect the changes that have/have not been made. @alelom please correct me if I am confabulating.
> 
> You kind of can if you modify the objects by reference, it's how the Adapter_Ids get's assigned to the objects output on the other side of the push. ETABS does some modifications on Bars, which are flipped sometimes, by keeping the reference and reflecting that change in the output, but as you mention it I'm unsure of the official stance on this? @alelom
> 
> Noticed that `Room Bounding` disappears when you changed the type of the column, which is why it was hard to find.
> 
> But tested through and happy with the PR.

Agreed with @kThorsager . As long as the objects are being cloned at the start, you are free to do whatever you want with them. The default push does this in the `ProcessObjectsForPush` method, if the adapter setting `CloneBeforePush` is set to true (which it is by default).

If you override the push, you can still call this method (which I think you already do) and if so, you can do whatever you want with the objects.

The only rule for the immutability is that the incoming objects/collection themselves can never be modified. As long as you have cloned them, you can do whatever you want with the clones, as you will pass this principle.

IsakNaslundBh:
Adding even further, I think the Revit adapter should be even more fine, for the general case where the sockets are being used, as sending the objects via sockets will implicitly clone them anyway, but good to make an explicit clone anyway for the case of the toolkit being used for dynamo for Revit.

pawelbaran:
Ah yeah, right, my bad - I was tired when typing the previous reply and got stuck in my wrong assumption. Nevertheless, tracking what had changed and what not would require a lot of overhead atm, therefore I would rather leave this for the next milestone when I will be revising the whole location/geometry-related concept.

PR #588, opened 2020-03-09
Unused files and legacy methods removed

### LINKED ISSUE #587: Revit_Toolkit: remove unused .cs files from the repo
#### Description:
As in the title, there is a few .cs files that do nothing floating around the repo, good to get rid of them.

pawelbaran:

Closes #587

### Test files

Rather not needed since the removed stuff was either not referenced in .csproj at all or the methods were never used (and I do not see use for them in reasonable future). Minor renaming sneaked in (__Properties_ turned back to _Properties_) should not require testing either.


pawelbaran:
@FraserGreenroyd @IsakNaslundBh could you please have a look at it any time soon? It should be a quick one and it starts to block further dev 😿

PR #659, opened 2020-04-21
CreateNonExsiting enabled

### LINKED ISSUE #629: Revit_Toolkit: support for PushType.CreateNonExisting
#### Description:
As in the title, following Adapter changes.

pawelbaran:

Closes #629

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue629%2DCreateNonExisting), follow the steps to test:
1. Open an empty Revit doc with BH structural template
2. Trigger first push from script
3. Delete some columns from Revit doc
4. Trigger second push from script
5. Inspect the Revit doc

### DIFF
--- a/Adapter_Revit_UI/AdapterActions/Push.cs
+++ b/Adapter_Revit_UI/AdapterActions/Push.cs
@@ -111,6 +111,11 @@ namespace BH.UI.Revit.Adapter
 
                 if (pushType == PushType.CreateOnly)
                     pushed = Create(objectsToPush, pushConfig);
+                else if (pushType == PushType.CreateNonExisting)
+                {
+                    IEnumerable<IBHoMObject> toCreate = objectsToPush.Where(x => x.Element(document) == null);
+                    pushed = Create(toCreate, pushConfig);
+                }
                 else if (pushType == PushType.DeleteThenCreate)
                 {
                     List<IBHoMObject> toCreate = new List<IBHoMObject>();


kThorsager:
Testing as instructed worked fine, but if one turned of the second one, deleted some more and re enabled it, it tried to create those it created last time again.
And restarting the adapter did not help.

Not even using the output from the second push as the input in my third push helped. On a second look that is as that only returns the objects which were created, but I think that is why that push always returns false (have a vague memory from the adapter about comparing counts)

I'm also getting the strong suspicion that this is based entirely on the id assigned on the first push, as attempting to use create non-existing on some un-pushed elements failed miserably. 

So, can one only use this once per script, and could we get some warnings and such

pawelbaran:
@kThorsager thanks for giving it a challenge. To put a bit of structure into the thread:

In general, Revit elements are compared based on their ElementId, as explained in the [Wiki](https://github.com/BHoM/Revit_Toolkit/wiki/Push-types).

> Testing as instructed worked fine, but if one turned of the second one, deleted some more and re enabled it, it tried to create those it created last time again.

This makes perfect sense: you were still trying to re-push all 100 items from the 1st step, so it placed the elements back in Revit.

> Not even using the output from the second push as the input in my third push helped.

Helped with what? Not sure if I understand. The ElementIds of elements created in the second push are different from the ones created in the first push as they are newly created, consequently their correspondent BHoM objects also carry new ElementIds:
image


> On a second look that is as that only returns the objects which were created, but I think that is why that push always returns false (have a vague memory from the adapter about comparing counts)

Yeah that is how it works, only the elements that have actually been pushed are returned - that is my understanding of what should be returned. By the way, in my opinion there will almost never be a definite true/false result of Push action - what if 99 of 100 objects get pushed correctly, is it true or false? Etc. @alelom 

> I'm also getting the strong suspicion that this is based entirely on the id assigned on the first push, as attempting to use create non-existing on some un-pushed elements failed miserably.

As above, each element successfully pushed in each step has a new ElementId (it gets assigned on create in Revit and cannot be driven by the user/API by any means) - same occurs to correspondent BHoM objects.

Not sure what do you mean by un-pushing in _ attempting to use create non-existing on some un-pushed elements_, so hard to answer to that.

> So, can one only use this once per script, and could we get some warnings and such?

One can use `CreateOnly` as many times as needed, but one has to be aware of ElementId matching mechanism. I think this is quite logical and covered in the Wiki - when thinking about it now, I am not sure if much more can be done to that...

When would you expect the warning to appear?

pawelbaran:
Thanks @kThorsager. And yeah, comparing actual elements would be way better, but it will take long until we get there I am afraid (as long as Autodesk does not expose its in-built diffing engine which is pretty powerful).

In general, I am aware that current Push is faaar from perfect, so any critics, suggestions, thoughts or recommendations are welcome.

Issue #664, opened 2020-04-21
Revit_Toolkit: CustomData key Revit_worksetId is duplicate of Workset


pawelbaran:
#### Description:

Both `CustomData` keys store the same information - this 1. is confusing 2. can cause issues on `SetProperty`. Therefore, it will be good to get rid of one, _Revit_worksetId_ being first to go as it is artificial and less user friendly.

Issue #638, opened 2020-03-30
Revit_Toolkit: Review and unify material converts


pawelbaran:
#### Description:

Atm each discipline handles it slightly differently. To be unified in a clear and reproducible pattern. Important to allow linking material properties with instance parameters.

PR #675, opened 2020-04-23
CustomData handling cleanup vol. 2

### LINKED ISSUE #670: Revit_Toolkit: Do not copy discipline-related information to object's CustomData on convert
#### Description:

Discipline-related information useful anywhere beyond Revit should be stored as object's properties and not `CustomData`.

Examples:
https://github.com/BHoM/Revit_Toolkit/blob/7d16834d1ddb69de8debe768f829edcd3ea93ac1/Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs#L158-L160

https://github.com/BHoM/Revit_Toolkit/blob/7d16834d1ddb69de8debe768f829edcd3ea93ac1/Engine_Revit_UI/Convert/Environment/FromRevit/Building.cs#L102-L104

pawelbaran:

Closes #670

### Test files

[BEnv macro pull script](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2F%5FMacro%2FPull%2FEnvironment%2FFrom%20MajaLindroth) is possibly the best way to test, project scripts are also welcome.

### Changelog
- discipline-related information stopped from being copied overt to object's `CustomData` on convert
- `BH.UI.Revit.Engine.Query.WorksetId` method has been deleted as a follow up to #664

### DIFF
--- a/Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs
+++ b/Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs
@@ -155,9 +155,6 @@ namespace BH.UI.Revit.Engine
                 material.YoungsModulus = youngsModulus.X;
                 material.ThermalExpansionCoeff = thermalExpansionCoeff.X;
                 material.PoissonsRatio = poissonsRatio.X;
-                material.CustomData["Concrete Bending Reinforcement"] = structuralAsset.ConcreteBendingReinforcement;
-                material.CustomData["Concrete Shear Reinforcement"] = structuralAsset.ConcreteShearReinforcement;
-                material.CustomData["Concrete Shear Strength Reduction"] = structuralAsset.ConcreteShearStrengthReduction;
                 materialFragment = material;
             }
             else if (materialFragment is Steel)
--- a/Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs
+++ b/Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs
@@ -155,6 +155,9 @@ namespace BH.UI.Revit.Engine
                 material.YoungsModulus = youngsModulus.X;
                 material.ThermalExpansionCoeff = thermalExpansionCoeff.X;
                 material.PoissonsRatio = poissonsRatio.X;
+                material.CustomData["Concrete Bending Reinforcement"] = structuralAsset.ConcreteBendingReinforcement;
+                material.CustomData["Concrete Shear Reinforcement"] = structuralAsset.ConcreteShearReinforcement;
+                material.CustomData["Concrete Shear Strength Reduction"] = structuralAsset.ConcreteShearStrengthReduction;
                 materialFragment = material;
             }
             else if (materialFragment is Steel)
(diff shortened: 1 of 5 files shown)


Engine_Revit_UI/Convert/Structure/FromRevit/MaterialFragment.cs
```diff
@@ -155,9 +155,6 @@ private static IMaterialFragment Update(this IMaterialFragment materialFragment,
                 material.YoungsModulus = youngsModulus.X;
                 material.ThermalExpansionCoeff = thermalExpansionCoeff.X;
                 material.PoissonsRatio = poissonsRatio.X;
-                material.CustomData["Concrete Bending Reinforcement"] = structuralAsset.ConcreteBendingReinforcement;
-                material.CustomData["Concrete Shear Reinforcement"] = structuralAsset.ConcreteShearReinforcement;
-                material.CustomData["Concrete Shear Strength Reduction"] = structuralAsset.ConcreteShearStrengthReduction;
```
FraserGreenroyd, line 160: This change affects @IsakNaslundBh heavily - worth him checking this just in case.

pawelbaran: Good point, I have deleted that somehow automatically. My gut feeling is that it had never been used anywhere as it is a mere `CustomData` key, not a property, but good to wait for @IsakNaslundBh's opinion on that.

IsakNaslundBh: From the side of the core BHoM, removing this should be absolutely fine. No methods or similar rely on these custom data values being set (if we did, would have made it a core property of the concrete material).

Only risk I can think of is if someone has a project script set up that makes use of these parameters. I would expect that to be fairly unlikely, but can ofc not be 100% sure.

Just to check, why are they removed @pawelbaran ? Not saying it is wrong, but just to confirm.

pawelbaran: They are removed as the general notion set by @FraserGreenroyd is to avoid storing discipline-specific information in `CustomData`, assuming that if something is meant to be usable later on in the BHoM, it should have a property to be assigned to.

From now on, only parameter values are meant to be copied across to `CustomData` (they will ultimately be migrated to `Parameters` fragment).

pawelbaran: Does the above satisfy you @FraserGreenroyd @IsakNaslundBh?

IsakNaslundBh: It does, but considering if they are worth keeping for now, as we do not have a strategy in place for how to handle them yet. The reinforcement parameters will never be (or never say never, but...) part of the Concrete material in structures, as we will want to handle reinforcement more explicitly, but, you might maybe still be able to pull the rest of the data (as revit generally works).

So would probably argue, lets keep, until another strategy for how to deal with those particular parameters is in place.

pawelbaran: After giving it another look I noticed this problem is likely to be much more convoluted - I have reverted the change in this PR and will rather challenge it with a wider slash in https://github.com/BHoM/Revit_Toolkit/issues/638. I believe this resolves the comment.

PR #707, opened 2020-05-15
Update of worksets enabled

### LINKED ISSUE #663: Revit_Toolkit: Workset cannot be set from CustomData
#### Description:

On push Revit parameters get updated based on `CustomData` - this generally works with one exception: workset, which does not work on change of _Workset_.

AndreBalasian:

Closes #663

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue663%2DWorkset%5Fchange&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### DIFF
--- a/Engine_Revit_UI/Create/Workset.cs
+++ b/Engine_Revit_UI/Create/Workset.cs
@@ -1,0 +1,88 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BH.UI.Revit.Engine
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        /// <summary>
+        /// Creates a new workset with a specified name and sets the "Workset" dropdown parameter to it.
+        /// </summary>
+        /// <param name="document"></param>
+        /// <param name="parameter"></param>
+        /// <param name="worksetName"></param>
+        /// <returns></returns>
+        public static bool Workset(Document document, Parameter parameter, string worksetName)
+        {
+            // Skip non-workshared documents because they can't contain worksets
+            if (document.IsWorkshared == false)
+                return false;
+
+            // Skip non-string workset names
+            if (worksetName is string == false)
+                return false;
+
+            // Skip null or empty workset names
+            if (string.IsNullOrWhiteSpace(worksetName))
+                return false;
+
+            // Skip non-workset parameters
+            if (parameter.Id.IntegerValue != (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+                return false;
+
+            // Skip null and read-only parameters
+            if (parameter == null || parameter.IsReadOnly)
+                return false;
+
+            // Avoid recreating a workset if one with the same name already exists
+            FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
+            Workset existingWorkset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
+            if (existingWorkset != null)
+            {
+                // Set the "Workset" dropdown parameter to the existing workset without recreating a new one
+                parameter.Set(existingWorkset.Id.IntegerValue); // Select a dropdown list item
+                return true;
+            }
+
+            // No existing worksets with the same name were found
+            // Create a new workset
+            Workset workset = null;
+            workset = Autodesk.Revit.DB.Workset.Create(document, worksetName); // Full namespace specified to avoid a name conflict with an Autodesk method
+
+            // Set the dropdown parameter to the newly created workset
+            parameter.Set(workset.Id.IntegerValue); // Select a dropdown list item
+
+            return true;
+        }
+
+        /***************************************************/
+    }
+}
+
--- a/Engine_Revit_UI/Create/Workset.cs
+++ b/Engine_Revit_UI/Create/Workset.cs
@@ -34,19 +34,18 @@ namespace BH.UI.Revit.Engine
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Creates a new workset with a specified name.")]
+        [Description("Creates a new workset if one does not exist under the specified name, otherwise returns the existing one.")]
         [Input("document", "Revit model file.")]
         [Input("worksetName", "Name of the new workset to be added.")]
         [Output("workset", "Existing workset if it exists or a newly created workset.")]
         public static Workset Workset(Document document, string worksetName)
         {
-            // Skip non-workshared documents because they can't contain worksets
-            if (document.IsWorkshared == false)
+            // Skip non-workshared documents and null/empty workset names
+            if (!document.IsWorkshared || string.IsNullOrWhiteSpace(worksetName))
+            {
+                BH.Engine.Reflection.Compute.RecordError("Document must be work shared, and workset name cannot be null or empty.");
                 return null;
-
-            // Skip null or empty workset names            
-            if (string.IsNullOrWhiteSpace(worksetName))
-                return null;
+            }
 
             // Avoid recreating a workset if one with the same name already exists
             FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
@@ -54,6 +53,7 @@ namespace BH.UI.Revit.Engine
             if (workset != null)
             {
                 // The specified workset already exists
+                BH.Engine.Reflection.Compute.RecordWarning("A workset with the same name already exists and is therefore reused.");
                 return workset; // Return the existing workset
             }
             else
--- a/Engine_Revit_UI/Create/Workset.cs
+++ b/Engine_Revit_UI/Create/Workset.cs
@@ -34,52 +34,36 @@ namespace BH.UI.Revit.Engine
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Creates a new workset with a specified name and sets the 'Workset' dropdown parameter to it.")]
-        [Input("document", "Revit model file")]
-        [Input("parameter", "Revit parameter")]
-        [Input("worksetName", "Name of the new workset")]
-        [Output("bool", "true for success, false for failure")]
-        public static bool Workset(Document document, Parameter parameter, string worksetName)
+        [Description("Creates a new workset with a specified name.")]
+        [Input("document", "Revit model file.")]
+        [Input("worksetName", "Name of the new workset to be added.")]
+        [Output("workset", "Existing workset if it exists or a newly created workset.")]
+        public static Workset Workset(Document document, string worksetName)
         {
             // Skip non-workshared documents because they can't contain worksets
             if (document.IsWorkshared == false)
-                return false;
+                return null;
 
-            // Skip non-string workset names
-            if (worksetName is string == false)
-                return false;
-
-            // Skip null or empty workset names
+            // Skip null or empty workset names            
             if (string.IsNullOrWhiteSpace(worksetName))
-                return false;
-
-            // Skip non-workset parameters
-            if (parameter.Id.IntegerValue != (int)BuiltInParameter.ELEM_PARTITION_PARAM)
-                return false;
-
-            // Skip null and read-only parameters
-            if (parameter == null || parameter.IsReadOnly)
-                return false;
+                return null;
 
             // Avoid recreating a workset if one with the same name already exists
             FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
-            Workset existingWorkset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
-            if (existingWorkset != null)
+            Workset workset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
+            if (workset != null)
             {
-                // Set the "Workset" dropdown parameter to the existing workset without recreating a new one
-                parameter.Set(existingWorkset.Id.IntegerValue); // Select a dropdown list item
-                return true;
+                // The specified workset already exists
+                return workset; // Return the existing workset
             }
+            else
+            {
+                // No existing workset with the same name was found
+                // Create a new workset
+                workset = Autodesk.Revit.DB.Workset.Create(document, worksetName); // Full namespace path specified to avoid a name clash with "BH.UI.Revit.Engine.Create", which cannot be omitted with "using Autodesk.Revit.DB;"
 
-            // No existing worksets with the same name were found
-            // Create a new workset
-            Workset workset = null;
-            workset = Autodesk.Revit.DB.Workset.Create(document, worksetName); // Full namespace specified to avoid a name conflict with an Autodesk method
-
-            // Set the dropdown parameter to the newly created workset
-            parameter.Set(workset.Id.IntegerValue); // Select a dropdown list item
-
-            return true;
+                return workset; // Return the newly created workset
+            }
         }
 
         /***************************************************/
--- a/Engine_Revit_UI/Create/Workset.cs
+++ b/Engine_Revit_UI/Create/Workset.cs
@@ -21,7 +21,9 @@
  */
 
 using Autodesk.Revit.DB;
+using BH.oM.Reflection.Attributes;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 
 namespace BH.UI.Revit.Engine
@@ -32,13 +34,11 @@ namespace BH.UI.Revit.Engine
         /****              Public methods               ****/
         /***************************************************/
 
-        /// <summary>
-        /// Creates a new workset with a specified name and sets the "Workset" dropdown parameter to it.
-        /// </summary>
-        /// <param name="document"></param>
-        /// <param name="parameter"></param>
-        /// <param name="worksetName"></param>
-        /// <returns></returns>
+        [Description("Creates a new workset with a specified name and sets the 'Workset' dropdown parameter to it.")]
+        [Input("document", "Revit model file")]
+        [Input("parameter", "Revit parameter")]
+        [Input("worksetName", "Name of the new workset")]
+        [Output("bool", "true for success, false for failure")]
         public static bool Workset(Document document, Parameter parameter, string worksetName)
         {
             // Skip non-workshared documents because they can't contain worksets
--- a/Engine_Revit_UI/Modify/SetParameter.cs
+++ b/Engine_Revit_UI/Modify/SetParameter.cs
@@ -133,38 +133,28 @@ namespace BH.UI.Revit.Engine
 
         public static bool SetParameter(this Parameter parameter, object value, Document document = null)
         {
-            // Skip non-workshared documents because they can't contain worksets
-            if (document.IsWorkshared == false)
-                return false;
-
-            // Skip non-string workset names
-            string worksetName = value as string;
-            if (worksetName is string == false)
-                return false;
-
-            // Skip null or empty workset names
-            if (string.IsNullOrWhiteSpace(worksetName))
-                return false;
-
-            // Skip null and read-only parameters
-            if (parameter == null || parameter.IsReadOnly)
-                return false;
-
             // Workset parameters
             // Filter for worksets
             if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
             {
-                FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
+                // Skip non-workshared documents because they can't contain worksets
+                if (document.IsWorkshared == false)
+                    return false;
 
-                Workset workset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
+                // Skip null or empty workset names
+                string worksetName = value as string;
+                if (string.IsNullOrWhiteSpace(worksetName))
+                    return false;
 
-                // Set the dropdown parameter to the specified workset if it exists
-                if (workset != null)
-                    return parameter.Set(workset.Id.IntegerValue);
+                // Skip null and read-only parameters
+                if (parameter == null || parameter.IsReadOnly)
+                    return false;
 
-                // Workset doesn't exist
-                // Create a new workset and set the "Workset" dropdown parameter to it
-                Create.Workset(document, parameter, worksetName);
+                // Create a new workset with a specified name or return an existing one if it already exists
+                Workset workset = Create.Workset(document, worksetName);
+
+                // Set the "Workset" parameter to the newly created workset
+                return parameter.Set(workset.Id.IntegerValue);
             }
 
             switch (parameter.StorageType)
--- a/Engine_Revit_UI/Modify/SetParameter.cs
+++ b/Engine_Revit_UI/Modify/SetParameter.cs
@@ -133,6 +133,19 @@ namespace BH.UI.Revit.Engine
 
         public static bool SetParameter(this Parameter parameter, object value, Document document = null)
         {
+            // Skip non-workshared documents because they can't contain worksets
+            if (document.IsWorkshared == false)
+                return false;
+
+            // Skip non-string workset names
+            string worksetName = value as string;
+            if (worksetName is string == false)
+                return false;
+
+            // Skip null or empty workset names
+            if (string.IsNullOrWhiteSpace(worksetName))
+                return false;
+
             // Skip null and read-only parameters
             if (parameter == null || parameter.IsReadOnly)
                 return false;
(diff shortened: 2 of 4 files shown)


Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+            // Skip null and read-only parameters
             if (parameter == null || parameter.IsReadOnly)
                 return false;
 
+            // Workset parameters
+            // Filter for worksets
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+            {
+                // Skip empty values
+                if (string.IsNullOrWhiteSpace(value as string))
```
FraserGreenroyd, line 145: What happens if `value` is not a `string`?

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                    return false;
+
+                // Find all user worksets
+                FilteredWorksetCollector worksets =
```
FraserGreenroyd, line 149: This could easily be a single line command

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                    new FilteredWorksetCollector(document)
+                    .OfKind(WorksetKind.UserWorkset);
+
+                // Does a workspace with the specified name exist?
+                foreach (Workset workset in worksets)
+                {
+                    if (workset.Name == value as string)
```
FraserGreenroyd, line 156: What happens if `value` is not a `string`?

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                if (string.IsNullOrWhiteSpace(value as string))
+                    return false;
+
+                // Find all user worksets
+                FilteredWorksetCollector worksets =
+                    new FilteredWorksetCollector(document)
+                    .OfKind(WorksetKind.UserWorkset);
+
+                // Does a workspace with the specified name exist?
+                foreach (Workset workset in worksets)
```
FraserGreenroyd, line 154: I would remove this and replace with something like the following - just less computationally expensive than a `for` loop going over all elements when one might not exist. Also less code to maintain/read :smile:

```
Workset workset = worksets.Where(x => x.Name == (value as string)).FirstOrDefault(); //Also need to check what happens if value is not a string

if(workset != null)
    return paremeter.Set(workset.Id.IntegerValue);
```

FraserGreenroyd:
> EDIT: apart from the appveyor things, but builds on my machine @FraserGreenroyd ?

Dunno why it broke but I hit it with a stick and it worked :smile:

Engine_Revit_UI/Modify/SetParameter.cs
pawelbaran: Good point, a check should happen at the beginning

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                    {
+                        // Workset found
+                        // Switch the dropdown menu to the required workset based on its index number
+                        int worksetIndex = workset.Id.IntegerValue; // Index number of the existing workset
+                        return parameter.Set(worksetIndex); // Select a dropdown list item
+                    }
+                }
+
+                // Workset doesn't exist
+                if (document.IsWorkshared) // Skip non-workshared documents
```
pawelbaran, line 166: This check should happen at the beginning

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,49 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                {
+                    // Create a new workset
+                    Workset newWorkset = null;
+                    string worksetName = value as string;
+                    newWorkset = Workset.Create(document, worksetName);
+
+                    // Select the newly created workset from the dropdown list
+                    int worksetIndex = newWorkset.Id.IntegerValue; // Index number of the new workset
+                    return parameter.Set(worksetIndex); // Select a dropdown list item
+                }
```
pawelbaran, line 176: After adding a few necessary checks, this could become a separate `Create` method

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Added type validation to the top

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Reduced to a single line

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Relocated to the top

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Reimplemented in Linq as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Moved to the top

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Migrated to a Create method as suggested with various checks upfront

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,88 @@
...
+        /****              Public methods               ****/
+        /***************************************************/
+
+        /// <summary>
+        /// Creates a new workset with a specified name and sets the "Workset" dropdown parameter to it.
+        /// </summary>
+        /// <param name="document"></param>
+        /// <param name="parameter"></param>
+        /// <param name="worksetName"></param>
+        /// <returns></returns>
```
kThorsager, line 41: Neat with descriptions but use 
`[Description("")]`
for compliance with the rest of the code

AndreBalasian: Reformatted the method description for compliance with the documentation convention

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,88 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Creates a new workset with a specified name and sets the 'Workset' dropdown parameter to it.")]
+        [Input("document", "Revit model file")]
+        [Input("parameter", "Revit parameter")]
+        [Input("worksetName", "Name of the new workset")]
+        [Output("bool", "true for success, false for failure")]
+        public static bool Workset(Document document, Parameter parameter, string worksetName)
```
pawelbaran, line 42: I believe this should take only `document` and `worksetName` variables and return a `Workset` type. Setting the parameter is a separate action to be taken in `SetParameters`.

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Creates a new workset with a specified name and sets the 'Workset' dropdown parameter to it.")]
+        [Input("document", "Revit model file")]
+        [Input("parameter", "Revit parameter")]
+        [Input("worksetName", "Name of the new workset")]
+        [Output("bool", "true for success, false for failure")]
```
pawelbaran, line 41: We are trying to be as neat and informative as possible when it comes to descriptions. This means:
- full sentences starting with capital and ending with a full stop
- reasonably rich information about arguments, e.g. "Name of the new workset to be added." instead of "Name of the new workset" etc.

Besides that, other remarks coming to my head:
- 'Workset' parameter is a dropdown only in Revit - on code level there is no dropdown 🙈 
- I would name the output "success" if it was a bool, but in case of returning `Workset` type I would probably  go for simple "workset"

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,40 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
         public static bool SetParameter(this Parameter parameter, object value, Document document = null)
         {
+            // Skip non-workshared documents because they can't contain worksets
+            if (document.IsWorkshared == false)
+                return false;
+
+            // Skip non-string workset names
+            string worksetName = value as string;
+            if (worksetName is string == false)
+                return false;
```
pawelbaran, line 143: these lines are redundant

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,40 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+
+            // Skip null or empty workset names
+            if (string.IsNullOrWhiteSpace(worksetName))
+                return false;
```
pawelbaran, line 147: All that code should be placed inside `if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)` - it is irrelevant if we are not trying to attempt _Workset_ parameter

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,40 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
             if (parameter == null || parameter.IsReadOnly)
                 return false;
 
+            // Workset parameters
+            // Filter for worksets
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+            {
+                FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
+
+                Workset workset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
```
pawelbaran, line 159: Here I would call `Create.Workset(document, worksetName)` method that would either return an existing workset with given name or a newly created one. Next step would be to set this workset to parameter.

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Revised accordingly

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Revised accordingly

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Removed accordingly

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Enclosed in the if statement as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Revised accordingly

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,72 @@
...
+        /***************************************************/
+
+        [Description("Creates a new workset with a specified name.")]
+        [Input("document", "Revit model file.")]
+        [Input("worksetName", "Name of the new workset to be added.")]
+        [Output("workset", "Existing workset if it exists or a newly created workset.")]
+        public static Workset Workset(Document document, string worksetName)
+        {
+            // Skip non-workshared documents because they can't contain worksets
+            if (document.IsWorkshared == false)
```
FraserGreenroyd, line 44: I would go for `if(!document.IsWorkshared)`

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,72 @@
...
+                return null;
```
FraserGreenroyd, line 45: Is it worth adding a `BH.Engine.Reflection.Compute.RecordError("Document cannot be workshared to create a Workset");` so that users know why they're getting a `null` result?

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            // Skip null or empty workset names            
+            if (string.IsNullOrWhiteSpace(worksetName))
+                return null;
```
FraserGreenroyd, line 49: I would simplify this to just be:

```
if(!document.IsWorkshared || string.IsNullOrWhiteSpace(worksetName))
{
    BH.Engine.Reflection.Compute.RecordError("Document must be work shared, and workset name cannot be null");
    return null;
}
```

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            // Avoid recreating a workset if one with the same name already exists
+            FilteredWorksetCollector worksets = new FilteredWorksetCollector(document).OfKind(WorksetKind.UserWorkset); // Find all user worksets
+            Workset workset = worksets.Where(x => x.Name == worksetName).FirstOrDefault(); // Find the specified workset if it exists
+            if (workset != null)
+            {
+                // The specified workset already exists
+                return workset; // Return the existing workset
```
FraserGreenroyd, line 57: Is it worth adding a warning to users to let them know they have been given an existing workset due to the name existing rather than a new one?

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,8 +133,29 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
         public static bool SetParameter(this Parameter parameter, object value, Document document = null)
         {
-            if (parameter == null || parameter.IsReadOnly)
-                return false;
+            // Workset parameters
+            // Filter for worksets
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+            {
+                // Skip non-workshared documents because they can't contain worksets
+                if (document.IsWorkshared == false)
```
FraserGreenroyd, line 141: I would simplify to `if(!document.IsWorkshared)`

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,8 +133,29 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
 
         public static bool SetParameter(this Parameter parameter, object value, Document document = null)
         {
-            if (parameter == null || parameter.IsReadOnly)
-                return false;
+            // Workset parameters
+            // Filter for worksets
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
```
FraserGreenroyd, line 138: If `parameter` is `null` on the entry point then this will throw a `NullReferenceException`.

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,8 +133,29 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                    return false;
+
+                // Skip null or empty workset names
+                string worksetName = value as string;
+                if (string.IsNullOrWhiteSpace(worksetName))
+                    return false;
+
+                // Skip null and read-only parameters
+                if (parameter == null || parameter.IsReadOnly)
+                    return false;
```
FraserGreenroyd, line 151: This should sit at the top of the method before this `if` block to protect if `parameter is null`.

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,8 +133,29 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+            {
+                // Skip non-workshared documents because they can't contain worksets
+                if (document.IsWorkshared == false)
+                    return false;
+
+                // Skip null or empty workset names
+                string worksetName = value as string;
+                if (string.IsNullOrWhiteSpace(worksetName))
+                    return false;
```
pawelbaran, line 147: This could be skipped here as same checks are made on `Create`

Engine_Revit_UI/Create/Workset.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace BH.UI.Revit.Engine
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Creates a new workset with a specified name.")]
```
pawelbaran, line 37: "Creates a new workset if one does not exist under specified name, otherwise returns the existing one."

AndreBalasian: Revised as suggested

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Revised as suggested

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Error message added

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Simplification implemented as suggested

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Warning added as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Null check moved up as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Simplified as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Removed as suggested

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Null check moved up as suggested

Engine_Revit_UI/Create/Workset.cs
pawelbaran: I would disagree with @FraserGreenroyd here. This warning will trigger every time an object is pushed/updated in Revit. I would rather add a warning when new workset is added - worksets are a very sensitive aspect of the model and are usually managed by the BIM manager. Adding a few by coincidence is the last thing that should happen, while using the existing workset is perfectly fine.

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,8 +133,25 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                // Skip non-workshared documents because they can't contain worksets
+                if (!document.IsWorkshared)
+                    return false;
+
+                // Create a new workset with a specified name or return an existing one if it already exists
+                string worksetName = value as string;
+                Workset workset = Create.Workset(document, worksetName);
+
+                // Set the "Workset" parameter to the newly created workset
+                return parameter.Set(workset.Id.IntegerValue);
```
pawelbaran, line 153: I would add an extra check here - by the read of code this may crash in case of whitespace workset name being fed in.

```suggestion
if (workset != null)
    return parameter.Set(workset.Id.IntegerValue);
else
    return false;
```

Engine_Revit_UI/Create/Workset.cs
FraserGreenroyd: Hmm by that logic @pawelbaran this wouldn't be a create method, if create is the edge case that we want to avoid? I would imagine a user (I would be at least) being confused if they used a create method and got a warning saying "hey, you just created an object, are you sure?" - I'd be like "yeah, course I'm sure, I asked for it" as opposed to using a 'create' method to get an in-use workset.

Based on your comment @pawelbaran it might be better to be a Query method instead, looking up an existing workset and returning a new one if one does not exist?

Engine_Revit_UI/Modify/SetParameter.cs
pawelbaran: This is not needed here at all actually - checked inside the `Create` method

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Empty workset names are already checked for in the Create method and return a null.
Would a null check like this suffice?
```
if (workset != null)
    return parameter.Set(workset.Id.IntegerValue);
```

Engine_Revit_UI/Create/Workset.cs
pawelbaran: Hmm, I would say it is a `Create` method as such @FraserGreenroyd, assuming that `Query` should not change the document, but only query it. However, I do agree that recording warning on expected behaviour would be hm... eccentric 🐒

On the other hand, warning on creation of a new workset and no warning on reusing the existing one is definitely the desirable end-user behaviour. This means we have 2 use cases with 2 different warnings:
- Attempted creation of a workset from scratch using `Create` method - desired warning when the existing workset is reused
- This case, i.e. setting workset param and warning the user if that action required creation of a new one

They are pretty much contradicting. Now, thinking about that a bit more, I would dare to say that creation of worksets is such a rare and sensitive case (max. few times in the entire project life?) that I would rather block it from happening by coincidence (on push of a single element). So, I lean towards checking if a workset exists under a given name, and if not then warning and no action. `Create` method disappears then and gets reimplemlemented as `Convert` once we have a BHoM wrapper for it that could be pushed.

What do you think @AndreBalasian?

Engine_Revit_UI/Modify/SetParameter.cs
pawelbaran: Not really, because then the code will continue an empty run, even though we already know the outcome of this method will be `false`.

AndreBalasian: A final 'else' statement will be added to capture that and return a false

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: I suggest having both a Query and a Create method.
The Query method would be called first and return the required workset if it exists, or trigger a Create method if it doesn't.

AndreBalasian: As for the warning message, let's leave that out of the Query method and only produce a warning through the Create method when the Query method calls it.

pawelbaran: Again, I believe it is dangerous from the project point of view. After a deeper thought, I cannot imagine pushing worksets somewhat as an attachement to single elements - this seems to be completely against general Revit policy and BIM standards. Elements should be driven by worksets, not the opposite.

So to allow creating Revit worksets from BHoM, I would strongly insist on creating a wrapper in Revit_oM and correspondent `Convert` method in Engine_Revit_UI. By these means we will make sure that the worksets will get created only on purpose (which I believe should be the case).

Engine_Revit_UI/Modify/SetParameter.cs
AndreBalasian: Removed as suggested

Engine_Revit_UI/Create/Workset.cs
AndreBalasian: Following a chat with @pawelbaran the Create method has been converted to a Query method that returns an existing workset with a specified name if it exists and null if it doesn't.
No new worksets will be created by this routine if they don't already exist.

Engine_Revit_UI/Query/Workset.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Find and return an existing workset with a specified name or return null if it doesn't exist.")]
+        [Input("document", "Revit model file.")]
+        [Input("worksetName", "Name of the required workset.")]
```
pawelbaran, line 39: I would try to go a bit more specific here.

```suggestion
        [Input("worksetName", "Name of the requested workset as it is shown in the UI of Revit.")]
```

Engine_Revit_UI/Query/Workset.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Output("workset", "Existing workset with a specified name if it exists or null if it doesn't.")]
+        public static Workset Workset(Document document, string worksetName)
```
pawelbaran, line 41: As a general principle, all BHoM `public static` methods are extension methods:

```suggestion
        public static Workset Workset(this Document document, string worksetName)
```

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,33 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
             if (parameter == null || parameter.IsReadOnly)
                 return false;
 
+            // Workset parameters
+            // Filter for worksets
+            if (parameter.Id.IntegerValue == (int)BuiltInParameter.ELEM_PARTITION_PARAM)
+            {
+                // Find an existing workset with a specified name if it exists
+                string worksetName = value as string;
+                Workset workset = Query.Workset(document, worksetName);
```
pawelbaran, line 146: As mentioned in the other comment, all `public static` methods in BHoM should be extension methods, so this could look as follows (and is preferred):

```suggestion
                Workset workset = document.Workset(worksetName);
```

Engine_Revit_UI/Modify/SetParameter.cs
```diff
@@ -133,9 +133,33 @@ public static bool SetParameter(this Element element, BuiltInParameter builtInPa
...
+                // Ensure the Query method hasn't returned a null, which can happen if the document is not workshared or the workset name is empty
+                if (workset != null)
+                {
+                    // Set the parameter to a workset with the specified name if it exists
+                    return parameter.Set(workset.Id.IntegerValue);
+                }
+                else
+                {
+                    // A workset with the specified name doesn't exist
+                    BH.Engine.Reflection.Compute.RecordError("Cannot set the Workset parameter because a workset with the specified name doesn't exist.");
```
pawelbaran, line 158: I would change it to a warning - errors are reserved to cases when an issue occurs that results in wrong/no result. This is just a single parameter not being set for a non-breaking reason, so warning sounds more suitable I think.

```suggestion
                    BH.Engine.Reflection.Compute.RecordWarning("Cannot set the Workset parameter because a workset with the specified name doesn't exist.");
```

PR #768, opened 2020-06-19
Zero length column error on push eliminated

### LINKED ISSUE #767: Revit_Toolkit: column push with base/top offset issue
#### Description:

Apparently, there is a bug in Revit that under some combinations of base vs top offset values one needs to go first, otherwise Revit will force cancel whole Push action without a particular reason. For more see the test file.

#### Test file(s):

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue767%2DPushColumnOffsetBug&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

pawelbaran:

Closes #767

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue767%2DPushColumnOffsetBug&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### DIFF
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -123,9 +123,13 @@ namespace BH.Revit.Engine.Core
                 }
             }
 
+            // Make sure the top is above base, otherwise Revit will complain for no reason.
+            familyInstance.get_Parameter((BuiltInParameter.FAMILY_BASE_LEVEL_OFFSET_PARAM)).Set(-1e+3);
+            familyInstance.get_Parameter((BuiltInParameter.FAMILY_TOP_LEVEL_OFFSET_PARAM)).Set(1e+3);
+
             familyInstance.CopyParameters(framingElement, settings);
             familyInstance.SetLocation(framingElement, settings);
-
+            
             refObjects.AddOrReplace(framingElement, familyInstance);
             return familyInstance;
         }
--- a/Revit_Core_Engine/Modify/CopyParameters.cs
+++ b/Revit_Core_Engine/Modify/CopyParameters.cs
@@ -91,6 +91,7 @@ namespace BH.Revit.Engine.Core
 
             if (propertyInfos != null)
             {
+                propertyInfos = propertyInfos.Where(x => x.PropertyType != typeof(double)).Union(propertyInfos.Where(x => x.PropertyType == typeof(double)));
                 foreach (PropertyInfo pInfo in propertyInfos)
                 {
                     object value = pInfo.GetValue(bHoMObject);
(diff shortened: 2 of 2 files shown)


Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -123,9 +123,13 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
                 }
             }
 
+            // Make sure the top is above base, otherwise Revit will complain for no reason.
+            familyInstance.get_Parameter((BuiltInParameter.FAMILY_BASE_LEVEL_OFFSET_PARAM)).Set(-1e+3);
+            familyInstance.get_Parameter((BuiltInParameter.FAMILY_TOP_LEVEL_OFFSET_PARAM)).Set(1e+3);
```
IsakNaslundBh, line 128: This seem to resolve the problem, and I think I get why you are doing this, but a bit scary to have this sort of hack in here, just to be overwritten below. Things that worry me a bit is that, what if for example we decide to refactor the `SetLocation` we might end up with columns that always have an offset of a total of 2000 feet (assume that is the standard unit).

Could this happen in `SetLocation` instead, or would that just lead to a bigger world of trouble?

pawelbaran: I do not like the solution either. However, it is the only solution I have found that fixes the issue. The problem is that Revit will keep on raising the error (and consequently taking back the whole push) every time the top of column moves above its base at _any_ point of parameter assignment, i.e. for the following case:
- base at offset +3000 from level +2000
- top at offset +3000 from level +4000

It will work if the top offset gets assigned first (order of assignment driven by the list order inside `RevitParametersToPush` fragment), but will fail if bottom goes first (after assigning bottom and before assigning top, bottom == 5000 > top == 4000).

The above can actually be tested by reversing the order of parameters in the test file - the column will get created then!

By stretching the column up and down I make sure the top will always be above the base. This does not too much sense but that is how Revit works ([I am not the only one struggling](https://forums.autodesk.com/t5/revit-api-forum/column-creation-questoin/m-p/2097225/highlight/true#M662)).

An alternative could be to check the elevation of the levels first, then manipulate the order of parameters to be changed. This, however, becomes a bit more tricky when parameter mapping needs to be taken into account. Plus it is also a hack 🙈 Therefore, taken the available time vs potential benefits, I decided to crowbar this issue a little bit.

pawelbaran:
Someone needs to keep the track in the end 🤣

PR #774, opened 2020-06-30
MaterialTakeoff implemented

### LINKED ISSUE #718: Revit_Toolkit: Implement RevitMaterialTakeoff fragment
#### Description:

It would be nice to add a Fragment that would contain a material takeoff information from Revit. Getting all materials from a given Revit element is simple combining `element.GetMaterialIds(true)` with `element.GetMaterialVolume()`.

@enarhi @kayleighhoude FYI

pawelbaran:

Closes #718

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue718%2DMaterialTakeoff&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### Changelog

- `RevitMaterialTakeoff` fragment has been implemented
- Scraping of material takeoff has been implemented to happen on every pull of `Physical` elements

### Additional comments

This PR is a fallout from a few discussions that I have listened to recently, that were focusing on LCA analysis - it is a quick POC that def needs more work
- empty `Materials` with only names are created instead of proper converts
- it is based on the current implementation of `MaterialComposition`, which seems not to make too much sense in this case (the volumes per material are wrapped into `TotalVolume` and `MaterialComposition` that operates on ratios, not very intuitive) - @al-fisher @kThorsager would be nice if you could have a look and maybe think whether `MaterialComposition` could not work directly on volumes instead of ratios
- missing descriptions
- will crash for zero materials due to bug in the construction of `MaterialComposition`

Before more work is put into this, I would like to hear the opinions of @kayleighhoude @enarhi @FraserGreenroyd whether it is helpful by any means, and if so, it would be good to map out how the classes should be structured.


kThorsager:
> ### Additional comments
> This PR is a fallout from a few discussions that I have listened to recently, that were focusing on LCA analysis - it is a quick POC that def needs more work
> 
> * empty `Materials` with only names are created instead of proper converts
> * it is based on the current implementation of `MaterialComposition`, which seems not to make too much sense in this case (the volumes per material are wrapped into `TotalVolume` and `MaterialComposition` that operates on ratios, not very intuitive) - @al-fisher @kThorsager would be nice if you could have a look and maybe think whether `MaterialComposition` could not work directly on volumes instead of ratios

Tried to scratch my head around this and as I remember it it was mostly of avoiding both necessitating knowing the total volume of something to create a MaterialComposition and making sure that if one combines two different ones they are in the same "scale" (that is one might assume that all have the total volume of each material, but one of them might be ratios)

But I have not made much use of or worked with it and am open for changes to reflect on the uses, which I think those in LCA know more of.

enarhi:
As far as if this is worth pushing further, I would say absolutely yes. The ability to take advantage of the MaterialTakeoff api method is huge, and is something that is currently really limited (even within Revit!). Other options for querying volumes of elements from Revit, even if implemented as IElementM, do not account for overlapping volumes and other things like that that the MaterialTakeoff can also account for. That is all to say that this is a very valuable feature add for the Revit Toolkit that I would anticipate would be taken advantage of for many workflows (LCA absolutely being one of them).

enarhi:
Test script I am using below. A couple notes:

- As shown, the material takeoff gives us volumes accounting for overlap, while also accounting for beam cutbacks. The cutback part is good but needs noting, as this similar volume accounting will not exist when using volume takeoffs from analysis programs or others. 
- Unfortunately Revit still does not play nice with composite deck accounting. We will need some other means to solve volumes for composite decks correctly, which has already proven to be a sticking point for the LCA workflow. 
- The total volume + composition output is not great but I get why it is currently set that way. I think the option Pawel notes for MaterialComposition working directly on volumes sounds ideal.

[MaterialTakeoffTest.zip](https://github.com/BHoM/Revit_Toolkit/files/4905175/MaterialTakeoffTest.zip)

pawelbaran:
@enarhi sounds good! The question is how to push that further - could you turn your notes into a list of requirements so that we could map out the next steps?

enarhi:
@pawelbaran sure thing, see below:

1. Accurate material takeoffs for all key single material structural element types populated on pulled elements as 'RevitMaterialTakeoff' fragment (already supported, possible tweak to fragment).
2. Accurate material takeoffs for decks accounting for deck profile to get real concrete volume. (Not supported by standard Revit MaterialTakeoff API method)
3.  'RevitMaterialTakeoff' to provide 'MaterialComposition' with Material and Volume so that we do not need to reverse engineer by ratios.

pawelbaran:
Thanks @enarhi. Is it possible to draw a formula that would cover 2.? Like area * thk of decking * corrugation factor, or similar?

@kThorsager is there a chance you could reflect on possibilities to refactor the `MaterialComposition` to make it a bit more user friendly? @al-fisher @LMarkowski FYI foundational

enarhi:
@pawelbaran 2 comes down to a combination of the deck's key dimensions, shown for example in this [Vulcraft Deck Brochure](https://vulcraft.com/upload/catalogs/en/decks/Vulcraft-Steel-Roof-Floor-Deck-Manual-Jun2018.pdf). You could use depth, P,T,and B from page 5 to get the concrete volume between flutes, then add that to the concrete volume above deck to get the real concrete volume. This is obviously a bit tricky coming from something like Revit, but Revit does provide those key parameters as well in a slightly different form, as shown below. The challenge is that we rarely model a deck profile exactly the same as the ones that will eventually be used. With that said, I think if Revit Toolkit returns the values based on the profile used in Revit thats great; we can resolve the need for greater profile detail outside of Revit_Toolkit.
image

pawelbaran:
Right, in this case it would be nice to convert the Revit deck information into a formula I guess? Is that something you could challenge @enarhi?

enarhi:
Yea, Revit deck info into a formula should do the trick. In this case:


 where

enarhi:
@pawelbaran I'm leaving the previous comment up but on second thought we should be doing the materialTakeoff multiplied by an effective thickness factor that is total effective thickness over thickness. So I think it would be better to implement:

kThorsager:
Only scannig throug your comments in general, but would the implementation for `AverageThickness` in Structure_Engine do the trick?
Looks like it's equivillant to @enarhi last comment, and is what is used to get the volume of ribbed structural slabs.
https://github.com/BHoM/BHoM_Engine/blob/324f5930f34f5dcf8926c5469f6355da86f75460/Structure_Engine/Query/AverageThickness.cs#L55

pawelbaran:
Thanks guys. So we have 2 questions to answer, as discussed above:
1. Whether to refactor `MaterialComposition`
2. How to capture the deck volume

Currently I am working on other more urgent issues, therefore will need to leave it hanging. However, feel free to step in @enarhi in case you want to push it a bit 👍

kThorsager:
> 1. Whether to refactor `MaterialComposition`

By popular demand it seems the answer is yes,

And as that would mean weaving in the volume into the `MaterialComposition` a name change might be good?

But more importantly `IElementM`s two extension methods `SolidVolume()` and `MaterialComposition()` should presumably be made into one as well. Since they would overlap in purpose.

And the existing methods should ofc be checked for where people have made assumptions about ratios being ratios.

pawelbaran:
> By popular demand it seems the answer is yes,

I can see the use of `MaterialComposition` without volumes (e.g. inside a definition of a cross-section) etc., so not sure if it is desirable to force it to materialize by linking it to volume. So would consider keeping it and maybe adding a new object to the oM, possibly an `IImmutable` that would link the composition with volume of each material.

Another issue are the UI limitations, i.e. how to store/expose the subvolumes to the user? We do not support dynamic numbers of properties nor generic dictionaries, therefore I guess this would need to be a `List` property containing wrapper objects  that would contain material and volume of each subvolume. This does not sound much more user friendly than what we have now.

Or am I getting confused?

kThorsager:
> I can see the use of `MaterialComposition` without volumes (e.g. inside a definition of a cross-section) etc., so not sure if it is desirable to force it to materialize by linking it to volume. So would consider keeping it and maybe adding a new object to the oM, possibly an `IImmutable` that would link the composition with volume of each material.

Excellent point.

> Another issue are the UI limitations, i.e. how to store/expose the subvolumes to the user? We do not support dynamic numbers of properties nor generic dictionaries, therefore I guess this would need to be a `List` property containing wrapper objects  that would contain material and volume of each subvolume. This does not sound much more user friendly than what we have now.

As of now, the `IImmutable` state of it means that we store it and two related collections. and doing as you say and storing them as a tuple or similar would probably not give much benefit.
The Materials are suppose to be unique, which draws one towards a Dictionary, but unsure on how to force it to use our comparer to make them properly unique, and as you say the UI limitations with that approach.

And think I did it that way as in the UI one would presumably do something for each pair of material and ratio, and that slotted right in with Grasshoppers treatment of lists. But actual use-cases are welcome.
And would say that the Ratio one and the Volume one should be similar in the UI at least.

al-fisher:
> I can see the use of MaterialComposition without volumes (e.g. inside a definition of a cross-section) etc., so not sure if it is desirable to force it to materialize by linking it to volume. So would consider keeping it and maybe adding a new object to the oM, possibly an IImmutable that would link the composition with volume of each material.

Agreed with comments above - now we have a few good working workflows - good to map our potential simplification of the MaterialComposition - for some of the cases. 

As always - this comes down to what we information we can calculate from or store on the object, and conversly what do we want to take as input. 
Volume being one of those tricky ones - that can sometimes be derived, sometimes pulled from another software's calculation, or thirdly indeed sometimes needed to overridden, or include other factors. .

As always think we will need a bit of both, enabling ratios and/or absolute. 
Be good for us to diagram up the next iteration - based on the two main contrasting workflows

Revit_oM/Revit_oM.csproj
```diff
@@ -53,6 +53,10 @@
       <HintPath>C:\ProgramData\BHoM\Assemblies\Geometry_oM.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Physical_oM">
+      <HintPath>..\..\..\..\..\..\..\ProgramData\BHoM\Assemblies\Physical_oM.dll</HintPath>
```
AndreBalasian, line 57: Minor corrections to the path

Revit_Engine/Query/RevitMaterialTakeoff.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace BH.Engine.Adapters.Revit
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("")]
```
AndreBalasian, line 37: Please populate the attributes

pawelbaran:
Closing this, PR reopened under #917

PR #791, opened 2020-07-06
Added ability to get identifier and type params using GetRevitParamValue

### LINKED ISSUE #789: Revit_Toolkit: enable querying RevitIdentifiers with GetRevitParameterValue
#### Description:

Currently properties of `RevitIdentifiers` cannot be queried by `GetRevitParameterValue` - would be nice to allow that. One thing to note, names of the properties should be somehow mapped with parameter names in Revit, preferably not by hardcoding, but e.g. by using same names for Revit parameters and properties of `RevitIdentifiers` type.

### LINKED ISSUE #790: Revit_Toolkit: make GetRevitParameterValue look for type parameters
#### Description:

Currently `GetRevitParameterValue` parses only the `RevitPulledParameters` fragment attached to the BHoM object representing the correspondent instance. It would be nice to allow parsing the `RevitPulledParameters` fragment attached to BHoM object's property that represent the Revit element type.

The challenge here is the fact that the BHoM object does not carry the information about which of its properties is correspondent to the type of Revit element. The easiest way to resolve that is to crowbar it via reflection (parse every property for `BHoMObjects` that have `RevitPulledParameters` fragment) - neater solutions are welcome 🤣

enarhi:
Closes #789 
Closes #790 

The GetRevitParamValue method can now get identifier and type parameters from objects pulled from Revit, which is a key feature to allow for downstream workflows.

### Test files
[RevitGetParamValueTest.zip](https://github.com/BHoM/Revit_Toolkit/files/4881605/RevitGetParamValueTest.zip)

### DIFF
--- a/Revit_Engine/Query/GetRevitParameterValue.cs
+++ b/Revit_Engine/Query/GetRevitParameterValue.cs
@@ -59,6 +59,14 @@ namespace BH.Engine.Adapters.Revit
                     return param.Value;
             }
 
+            RevitIdentifiers identifierFragment = bHoMObject.Fragments?.FirstOrDefault(x => x is RevitIdentifiers) as RevitIdentifiers;
+            if (identifierFragment != null)
+            {
+                string paramName = string.Concat(parameterName.Where(c => !char.IsWhiteSpace(c)));
+                if (Reflection.Query.PropertyNames(identifierFragment).Contains(paramName))
+                    return Reflection.Query.PropertyValue(identifierFragment, paramName);
+            }
+
             return null;
         }
 
--- a/Revit_Engine/Query/GetRevitParameterValue.cs
+++ b/Revit_Engine/Query/GetRevitParameterValue.cs
@@ -68,16 +68,22 @@ namespace BH.Engine.Adapters.Revit
                     return Reflection.Query.PropertyValue(identifierFragment, paramName);
             }
 
-            List<object> bHoMPropList = Reflection.Query.PropertyObjects(bHoMObject);
-            List<IBHoMObject> bHoMProps = bHoMPropList.OfType<IBHoMObject>().ToList();
-            foreach (IBHoMObject bHoMProp in bHoMProps)
+            Dictionary <string, object> bHoMPropDic = Reflection.Query.PropertyDictionary(bHoMObject);
+            foreach (KeyValuePair<string, object> bHoMPropEntry in bHoMPropDic)
             {
-                RevitPulledParameters typePullFragment = bHoMProp.Fragments?.FirstOrDefault(x => x is RevitPulledParameters) as RevitPulledParameters;
-                if (typePullFragment?.Parameters != null)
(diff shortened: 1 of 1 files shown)


enarhi:
@pawelbaran as you noted on the issue, this currently uses the "crowbar" solution of searching all IBHoMObject properties for a PulledParameters fragment. Performance looks to be acceptable (~100ms for 10000 Revit objects of various typical types) so I'm happy doing it that way if you are. Also worth noting as of now it is only searching for the type params on the PulledParams. Easy enough to also implement on PushParams but I am not sure how useful / necessary that may be?

Revit_Engine/Query/GetRevitParameterValue.cs
```diff
@@ -59,6 +60,27 @@ public static object GetRevitParameterValue(this IBHoMObject bHoMObject, string
...
+            List<object> bHoMPropList = Reflection.Query.PropertyObjects(bHoMObject);
+            List<IBHoMObject> bHoMProps = bHoMPropList.OfType<IBHoMObject>().ToList();
+            foreach (IBHoMObject bHoMProp in bHoMProps)
+            {
+                RevitPulledParameters typePullFragment = bHoMProp.Fragments?.FirstOrDefault(x => x is RevitPulledParameters) as RevitPulledParameters;
+                if (typePullFragment?.Parameters != null)
+                {
+                    RevitParameter param = typePullFragment.Parameters.FirstOrDefault(x => x.Name == parameterName);
+                    if (param != null)
+                        return param.Value;
```
pawelbaran, line 80: Could we please record a warning here saying that the value for an object under BHoM_Guid X has been retrieved from parameters belonging to its property Y?

pawelbaran:
> Also worth noting as of now it is only searching for the type params on the PulledParams. Easy enough to also implement on PushParams but I am not sure how useful / necessary that may be?

Yeah this sounds like a step too far to me tbh. However, what I would consider:
- `GetRevitParameterValue` with parameter name argument - parse, in the following order (exactly what the code in this PR does):
    - object's parameters to push
    - object's pulled parameters
    - object's pulled identifiers
    - object type's pulled parameters
- GetRevitParameterValue` with parameter name argument preceded by _Type:_ prefix:
    - object type's parameters to push
    - object type's pulled parameters
    - object type's pulled identifiers

Same approach could also be applied to Set methods. Do you think that would be useful in practice @enarhi?

kayleighhoude:
Looks good to me, appreciate the lack of prefix required for type parameters ATM: 
image

enarhi:
@pawelbaran done, warning now recorded when the type param is retrieved. Regarding the type parameter specific retrieval, I think that could be useful but is less urgent. It is somewhat rare that we need a type param specifically that the current methods would not allow for, and if the type and instance params exist with the same name they would tend to have the same value as well (something like "Comments" being a possible exception).

enarhi:
@pawelbaran this should be ready for review and merge now.

pawelbaran:
Thanks guys, I've raised #797 as a followup

PR #818, opened 2020-07-21
Revit_Toolkit: Revit => BHoM material converts unified and cleaned up

### LINKED ISSUE #404: Material.Create looks suspicious
#### Description:
As discussed with @ZiolkowskiJakub, material creation method looks suspicious and may cause issues on push.
https://github.com/BHoM/Revit_Toolkit/blob/d740e15aa7ce002e49089dfadd73d7e74e5c688b/Engine_Revit_UI/Convert/Physical/ToRevit/Material.cs#L43

#### Test file(s):
Would be worth creating a test file in the field of material push/maching?

### LINKED ISSUE #638: Revit_Toolkit: Review and unify material converts

### LINKED ISSUE #640: Revit_Toolkit: Extend refObjects keys to capture instance-dependent materials
#### Description:
The result of material convert can depend on an instance. In order to avoid overhead of converting the material per element, the converted materials could be added to `refObjects`:
- `ElementId.IntegerValue.ToString()` if the material is not dependent on any parameter
- `ElementId.IntegerValue.ToString() + MapType(parameter.Name)` if the material is dependent on a parameter

### LINKED ISSUE #645: Revit_Toolkit: BH.UI.Revit.Engine.Convert.MaterialPropertiesFromRevit should become Query.MaterialProperties
#### Description:

As in the title, `MaterialProperties` could possibly fit better to `Query` than `Convert`. To be resolved together with #638.

pawelbaran:

Closes #404
Closes #638
Closes #640
Closes #645

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2F%5FMacro%2FPull%2FMaterials&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) - manual inspection is needed to verify whether the pulled materials are OK. Push/update part of the test script sholud be checked only for correctnedss of the warnings/errors as explained in _Additional comments_ paragraph.

### Changelog

- `EmptyMaterialFromRevit` and `UpdateMaterialProperties` were methods removed to simplify the workflow and make it similar to other `Convert` methods (single `FromRevit` method per each type, without creating 'empty' objects and then updating)
- material grade argument added to all converts that include conversion of the material to either Structural or Physical
- use of `refObjects` has been enabled for grade-specific converts 
- the code base related to Revit => BHoM material converts has been cleaned up

### Additional comments

There is no clear strategy with regard to materials in BHoM/Revit, therefore this PR should get out as much as could be done, while keeping the code clean and less convoluted than it was before. I have not found any bugs, but would be thankful for a critical review.

**What does not work:** there is no support for updating the Revit material assets based on BHoM material fragments on Push - the framework for it has been created in this PR (see `CopyCharacteristics` methods), but needs to be populated with reliable converts. That does not seem to be high priority, therefore will be left as is, with _missing convert_ warnings automatically raised.


pawelbaran:
Flagged it as a WIP for a second, gotta add a bit more to this PR as the further dev fits in nicely.

pawelbaran:
WIP label taken off, ready for review @FraserGreenroyd @LMarkowski 

For those unhappy about the size of this PR - again, it was hard to split, so I decided to simply batch it together. Please also note that a substantial part of the changes concerns renaming and migrating the code, so please try to be positive 🤣 And I promise it is the last such big refactor of converts!

pawelbaran:
Nice spot @IsakNaslundBh. I've investigated the issue and unfortunately it seems impossible to avoid it. Each Revit framing element family has a predefined structural material type setting (see the screenshot below), while the actual material is assigned as a property of the instance of that family.

image

With the material grade attached as a string parameter of the instance, we have 3 sources of information about the material of an element (as simple as Revit 🙊):
1. material type specified in a family (enum with very general values like concrete/steel/timber etc.)
2. material object being a property of a family instance (can be any object of type `Material` specified by the user)
3. string property of a family instance containing the information about material grade

The algorithm runs the following sequence:
- take material type of the family (1. above)
- take material grade (3.) and check if it exists in the library for type extracted in the previous step
    - if yes, return the fragment from the library
    - else convert the material object from 2. and return it

By these means, if the material type and grade match the library, convert is not needed and the curated library object is returned.

**TL;DR:**
While writing this post I came up with an extra check that could be added to resolve this issue: compare `MaterialCategory` property of the material from 2. above with material type from 1 - if they mismatch, then do not query the library and do the full convert.

Sorry for the lengthy answer, I am afraid it will not make too much sense for a non-Revit dev 🤣 But I will try to fix it tomorrow, stay tuned!

pawelbaran:
BTW, I have just extended the test file to test family materials other than concrete in order to make the check more representative.

pawelbaran:
The functionality has been updated to extract maximum information possible:
- material extracted from element, if not found then from default for its category, if not then from unknown material returned
- material grade taken into account on convert to structural fragment and added as a suffix to the name of the latter
- bugs found by @IsakNaslundBh should not occur any more

Text files have been updated to support reviewing the above.

I am also adding @enarhi as a second reviewer as this PR may affect the work on LCA.

pawelbaran:
Agreed @enarhi. Currently the code will break in case of an attempt to convert a Revit panel with more than 1 structural layer into structural BHoM one. That is the limitation of the BHoM itself as we do not have a definition for composite `SurfaceProperty`. @IsakNaslundBh please shout if I am confusing things.

IsakNaslundBh:
@pawelbaran you are not confusing things. Currently only have support for single material surface properties in structures.

PR #850, opened 2020-09-18
Profile extraction bugs fixed

### LINKED ISSUE #848: Revit_Toolkit: Name matching issues in profile extraction
#### Description:

A few issues related to dimension name matching appeared when during profile extraction tests. TBC in the PR.

### LINKED ISSUE #849: Revit_Toolkit: Non-symmetrical freeform framing profiles come out mirrored
#### Description:

As in the title, when pulling non-symmetrical framing profiles, they come out mirrored around local y in case they are retrieved directly from family's geometry. That happens because of wrong end profile being taken.

pawelbaran:

Closes #848
Closes #849

### Test files

#848 on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/Installers/Forms/AllItems.aspx?RootFolder=%2fsites%2fBHoM%2fInstallers%2fScripts%2fBuroHappold%5fBHoM%5fv3%2e3%2f0001%5fRevit%5fToolkit%2fPull&FolderCTID=0x012000181C071E8B6D1E438B59C87DA36B9302):
- _BuroHappold_BHoM_v3.3.beta_PullColumnProfiles_ for columns
- _BuroHappold_BHoM_v3.3.beta_PullFramingProfiles_ for framing

**Important:** to make the GH script work fine, Rhino document tolerance should be set to 1e-6 (usually 1e-3 by default).

To select top row of supported types, one can use a predefined selection set named _1st row_:
image

#849 can be tested using the 2nd file from the above, taking one of the light gauge profiles, e.g. the one under ID 1388030.

### Changelog

- Name matching of profile dimensions improved
- Pull of freeform framing profile fixed to return the profile in correct orientation

### Additional comments

There is still a few bugs here and there, but as discussed with @IsakNaslundBh, it looks like we may not be able to fix all of them with the current approach - possible adjustments (cross checks or determination of profile shape from its outline) to be discussed on the Structures planning for 4.0. Following that, I am happy to merge the PR in its current state regardless the bugs.

### DIFF
--- a/Revit_Core_Engine/Convert/Structure/FromRevit/Profile.cs
+++ b/Revit_Core_Engine/Convert/Structure/FromRevit/Profile.cs
@@ -138,13 +138,13 @@ namespace BH.Revit.Engine.Core
             else if (section is StructuralSectionConcreteRound)
                 diameter = (section as StructuralSectionConcreteRound).Diameter.ToSI(UnitType.UT_Section_Dimension);
             else
-                diameter = familySymbol.LookupParameterDouble(diameterNames);
+                diameter = familySymbol.LookupParameterDouble(diameterNames, dimensionGroups);
 
             if (!double.IsNaN(diameter))
                 return BHG.Create.CircleProfile(diameter);
             else
             {
-                double radius = familySymbol.LookupParameterDouble(radiusNames);
+                double radius = familySymbol.LookupParameterDouble(radiusNames, dimensionGroups);
                 if (!double.IsNaN(radius))
                     return BHG.Create.CircleProfile(radius * 2);
             }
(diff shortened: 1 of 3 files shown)


LMarkowski (dismissed):
Code looks good.
Functionality tested on Rvt 2020. There are some minor bugs as annotated and discussed.
Approving basing on discussion.

Revit_Core_Engine/Convert/Structure/FromRevit/Profile.cs
```diff
@@ -171,18 +174,19 @@ private static FabricatedISectionProfile FabricatedISectionProfileFromRevit(this
...
+                botFlangeThickness = familySymbol.LookupParameterDouble(botFlangeThicknessNames, dimensionGroups);
+                weldSize = familySymbol.LookupParameterDouble(weldSizeNames1, dimensionGroups);
             }
 
             if (double.IsNaN(weldSize))
             {
-                weldSize = familySymbol.LookupParameterDouble(weldSizeNames2);
+                weldSize = familySymbol.LookupParameterDouble(weldSizeNames2, dimensionGroups);
                 if (!double.IsNaN(weldSize) && !double.IsNaN(webThickness))
                     weldSize = (weldSize - webThickness) / (Math.Sqrt(2));
```
IsakNaslundBh, line 191: Found in testing, think this should be:

```suggestion
                    weldSize = (weldSize - webThickness / 2) / (Math.Sqrt(2));
```

IsakNaslundBh (approved):
First of all, @pawelbaran massive compliments for the great test files. Would not have been able to pick up on the small fixes if it was not for them! 🚀 

Think this Generally works well, or well enough now. This PR is, for sure, a step in the right direction and gives far better results for various families than before.

As discussed, this is probably something we will need to come back to again to make it even more robust, but happy that this is a clear improvement that def should go into the beta release! Approved!

LMarkowski (approved):
Redone testing on RVT 2020 after requested changes. Everything works as good as before (or even better) so I'm upholding my review.

PR #906, opened 2020-11-16
Pulling nested elements implemented

### LINKED ISSUE #720: Revit_Toolkit: nested elements are not handled correctly
#### Description:

When pulling Revit elements with nested ones, only top level gets pulled - possibly nested elements should be pulled too.

#### Test file(s):

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FRevit%5FToolkit%2FRevit%5FToolkit%2DIssue720%2DNestedElements&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

enarhi:

Closes #720 

Nested elements are now pulled when pulling from Revit if the pull config option for IncludeNestedElements=True. Default behavior is currently to include nested elements; this can be changed in the future if default nested element pulls prove too heavy, although initial benchmarking indicates minimal extra overhead for most families with nested elements.

### Test files
[Revit and GH Test Files](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Revit_Toolkit/Revit_Toolkit-%23720-NestedElements?csf=1&web=1&e=kF7kkq)

### Additional comments
This PR includes implementation for nested instances in windows, which require unique methods due to how window geometry is calculated in revit. Most element types do not require this workaround, but some element types may going forward. These should be covered under separate issues on an as-needed basis.

See images below for examples of implementation, including storage of parent element id to a new revit identifier property. 

_Family with nested elems in Revit_
image

_Pulled in GH showing elem geo for nested elems_
image

_Props in Grasshopper per pull, including parent elem id. THis also shows the pull config with nested elements enabled._
image

### DIFF
--- a/Revit_Core_Adapter/CRUD/Read.cs
+++ b/Revit_Core_Adapter/CRUD/Read.cs
@@ -92,6 +92,23 @@ namespace BH.Revit.Adapter.Core
             if (representationConfig == null)
                 representationConfig = new PullRepresentationConfig();
 
+            if (pullConfig.IncludeNestedElements)
+            {
+                List<ElementId> elemIds = new List<ElementId>();
+                foreach (ElementId id in elementIds)
+                {
+                    elemIds.Add(id);
+                    Element element = document.GetElement(id);
+                    if (element is FamilyInstance)
+                    {
+                        FamilyInstance famInst = element as FamilyInstance;
+                        IEnumerable<ElementId> nestedElemIds = famInst.ElementIdsOfMemberElements();
+                        elemIds.AddRange(nestedElemIds);
+                    }
+                }
+                elementIds = elemIds;
+            }
+
             Options geometryOptions = BH.Revit.Engine.Core.Create.Options(ViewDetailLevel.Fine, geometryConfig.IncludeNonVisible, false);
             Options meshOptions = BH.Revit.Engine.Core.Create.Options(geometryConfig.MeshDetailLevel.ViewDetailLevel(), geometryConfig.IncludeNonVisible, false);
             Options renderMeshOptions = BH.Revit.Engine.Core.Create.Options(representationConfig.DetailLevel.ViewDetailLevel(), representationConfig.IncludeNonVisible, false);
--- a/Revit_Core_Adapter/CRUD/Read.cs
+++ b/Revit_Core_Adapter/CRUD/Read.cs
@@ -97,7 +97,6 @@ namespace BH.Revit.Adapter.Core
                 List<ElementId> elemIds = new List<ElementId>();
                 foreach (ElementId id in elementIds)
                 {
-                    elemIds.Add(id);
                     Element element = document.GetElement(id);
                     if (element is FamilyInstance)
                     {
@@ -106,7 +105,7 @@ namespace BH.Revit.Adapter.Core
                         elemIds.AddRange(nestedElemIds);
                     }
                 }
-                elementIds = elemIds;
+                elementIds.ToList<ElementId>().AddRange(elemIds);
             }
 
             Options geometryOptions = BH.Revit.Engine.Core.Create.Options(ViewDetailLevel.Fine, geometryConfig.IncludeNonVisible, false);
--- a/Revit_Core_Adapter/CRUD/Read.cs
+++ b/Revit_Core_Adapter/CRUD/Read.cs
@@ -74,7 +74,7 @@ namespace BH.Revit.Adapter.Core
             if (!pullConfig.IncludeClosedWorksets)
                 worksetPrefilter = document.ElementIdsByWorksets(document.OpenWorksetIds().Union(document.SystemWorksetIds()).ToList());
 
-            IEnumerable<ElementId> elementIds = request.IElementIds(uiDocument, worksetPrefilter).RemoveGridSegmentIds(document);
+            List<ElementId> elementIds = request.IElementIds(uiDocument, worksetPrefilter).RemoveGridSegmentIds(document).ToList<ElementId>();
             if (elementIds == null)
                 return new List<IBHoMObject>();
 
@@ -105,7 +105,7 @@ namespace BH.Revit.Adapter.Core
                         elemIds.AddRange(nestedElemIds);
                     }
                 }
-                elementIds.ToList<ElementId>().AddRange(elemIds);
+                elementIds.AddRange(elemIds);
             }
 
             Options geometryOptions = BH.Revit.Engine.Core.Create.Options(ViewDetailLevel.Fine, geometryConfig.IncludeNonVisible, false);
--- a/Revit_Core_Engine/Query/OpeningSurface.cs
+++ b/Revit_Core_Engine/Query/OpeningSurface.cs
@@ -21,7 +21,10 @@
  */
 
 using Autodesk.Revit.DB;
+using Autodesk.Revit.DB.IFC;
 using BH.Engine.Adapters.Revit;
+using BH.Engine.Geometry;
+using BH.oM.Adapters.Revit.Parameters;
 using BH.oM.Adapters.Revit.Settings;
 using BH.oM.Geometry;
 using System;
@@ -45,6 +48,12 @@ namespace BH.Revit.Engine.Core
             if (bbox == null)
                 return null;
 
+            // Get parent element if this family instance is a nested instance
+            Element parentElem = familyInstance.SuperComponent;
+            int parentId = -1;
+            if (parentElem != null)
+                parentId = familyInstance.SuperComponent.Id.IntegerValue;
+
             settings = settings.DefaultIfNull();
             List<HostObject> hosts;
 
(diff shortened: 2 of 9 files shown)


FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check code compliance

FraserGreenroyd:
@BHoMBot check copyright compliance

FraserGreenroyd (changes requested):
Please fix the documentation compliance issues to begin with :smile:

FraserGreenroyd:
@BHoMBot check project compliance

FraserGreenroyd:
@BHoMBot check project compliance

FraserGreenroyd (changes requested):
Also CSProject compliance please :smile:

enarhi:
@BHoMBot fix project file ref. 1409293750

FraserGreenroyd:
@BHoMBot check project compliance

FraserGreenroyd:
@BHoMBot check core

enarhi:
@BHoMBot check compliance

enarhi:
@BHoMBot check core

FraserGreenroyd:
@BHoMBot check core

Revit_Core_Engine/Query/OpeningSurface.cs
```diff
@@ -54,12 +63,102 @@ public static ISurface OpeningSurface(this FamilyInstance familyInstance, HostOb
...
+                FailureHandlingOptions failureHandlingOptions = t.GetFailureHandlingOptions().SetClearAfterRollback(true);
+                t.Start("Temp Delete Inserts");
+                List<ElementId> inserts = hosts.SelectMany(x => x.FindInserts(true, true, true, true)).Distinct().Where(x => x.IntegerValue != -1).ToList();
+                ParameterSet instanceParams = familyInstance.Parameters;
+                List<RevitParameter> bhomParams = new List<RevitParameter>();
+                foreach (Parameter parameter in instanceParams)
+                {
+                    RevitParameter bhomParam = Convert.ParameterFromRevit(parameter);
+                    bhomParams.Add(bhomParam);
+                }
```
pawelbaran, line 85: I do not understand this part of the code - why would you convert the Revit parameters to BHoM parameter objects and then do matching with hard coded names (_Sill Height_, _Head Height_)? This does not look sustainable and I would argue whether it is necessary.

Revit_Core_Engine/Query/OpeningSurface.cs
```diff
@@ -54,12 +63,102 @@ public static ISurface OpeningSurface(this FamilyInstance familyInstance, HostOb
...
+                    
+                LocationPoint locPt = familyInstance.Location as LocationPoint;
+                Element dummyInstance = doc.Create.NewFamilyInstance(locPt.Point, familyInstance.Symbol, hosts.First() as Element, Autodesk.Revit.DB.Structure.StructuralType.NonStructural);
```
pawelbaran, line 88: Why do you loop over all hosts if you place `dummyInstance` on the first? By the looks of code, you could use only it further down the line.

Revit_Core_Engine/Query/OpeningSurface.cs
```diff
@@ -54,12 +63,102 @@ public static ISurface OpeningSurface(this FamilyInstance familyInstance, HostOb
...
+                foreach (HostObject h in hosts)
+                {
+                    solidsWithOpening.Add(h.Solids(new Options()).Select(x => SolidUtils.Clone(x)).ToList());
+                }
+
+                // Rollback and restart of the transaction is needed because otherwise the object, to which familyInstance is pointing can become invalidated.
+                t.RollBack(failureHandlingOptions);
+                t.Start();
+
+                List<CurveLoop> loops = new List<CurveLoop>();
```
pawelbaran, line 110: This piece of code looks very similar to the existing implementation - couldn't it be wrapped into a private method and called twice to avoid boilerplate?

Revit_oM/Parameters/RevitIdentifiers.cs
```diff
@@ -55,12 +55,15 @@ public class RevitIdentifiers : IObject, IFragment, IImmutable
         [Description("ElementId of view that owns the Revit element correspondent to the BHoM object that carries this fragment. -1 if the Revit element is not view-dependent.")]
         public virtual int OwnerViewId { get; } = -1;
 
+        [Description("ElementId of the parent Revit element correspondent to the BHoM object that owns this.")]
+        public virtual int ParentElementId { get; } = -1;
```
pawelbaran, line 59: `ParentElementId` sounds a bit ambiguous to me - what is a parent? I guess people can understand it in various ways: a host, owner view, parent family etc. I would opt for something more precise, like, hm `SuperComponentId`? Or would that be too abstract?

Revit_oM/Parameters/RevitIdentifiers.cs
```diff
@@ -55,12 +55,15 @@ public class RevitIdentifiers : IObject, IFragment, IImmutable
...
+
 
         /***************************************************/
         /****            Public Constructors            ****/
         /***************************************************/
 
-        public RevitIdentifiers(string uniqueId = "", int elementId = -1, string categoryName = "", string familyName = "", string familyTypeName = "", int familyTypeId = -1, int ownerViewId = -1)
+        public RevitIdentifiers(string uniqueId = "", int elementId = -1, string categoryName = "", string familyName = "", string familyTypeName = "", int familyTypeId = -1, int ownerViewId = -1, int parentElementId = -1)
```
pawelbaran, line 66: One thing to notice, `RevitIdentifiers` implements `IImmutable` interface, which means it needs to be versioned individually in Versioning_Toolkit - see https://github.com/BHoM/Versioning_Toolkit/pull/104 as an example.

Revit_Engine/Create/Config/RevitPullConfig.cs
```diff
@@ -46,6 +46,20 @@ public static RevitPullConfig RevitPullConfig(Discipline discipline = Discipline
...
         /***************************************************/
 
+        [Description("Creates a pull action-specific configuration used for adapter interaction with Revit.")]
+        [InputFromProperty("discipline")]
+        [InputFromProperty("includeClosedWorksets")]
+        [InputFromProperty("geometryConfig")]
+        [InputFromProperty("representationConfig")]
+        [InputFromProperty("includeNestedElements")]
+        [Output("revitPullConfig")]
+        public static RevitPullConfig RevitPullConfig(Discipline discipline = Discipline.Undefined, bool includeClosedWorksets = false, PullGeometryConfig geometryConfig = null, PullRepresentationConfig representationConfig = null, bool includeNestedElements = true)
```
pawelbaran, line 56: Please deprecate the previous version of the method (possibly `PreviousVersionAttribute` could be used, too).

pawelbaran (changes requested):
I did a first round of code review - please see the code comments.

Functionality wise, I have tested it on pile caps and seems to run nicely 👍 However, it would be great if you could provide a few test cases @enarhi so that we could gather them together and build representative test content (windows, doors, pile caps, any other major type that is relevant to that issue).

Revit_oM/Parameters/RevitIdentifiers.cs
enarhi: noted, will add a pr to versioning accordingly once this is merged up

pawelbaran: Coordination will be welcome, otherwise all internalized objects pulled from Revit will change to `BHoMObjects` (that is how I learnt about it 😃).

Revit_oM/Parameters/RevitIdentifiers.cs
enarhi: Parent / Child adheres to standard nesting naming conventions. I think that is the most appropriate name for it, we could go with HostElement as well but that creates ambiguity between a Host is the nested sense and host in the revit host element sense (eg a wall hosting a window).

Revit_Core_Adapter/CRUD/Read.cs
```diff
@@ -92,6 +92,23 @@ protected override IEnumerable<IBHoMObject> Read(IRequest request, ActionConfig
...
+                    elemIds.Add(id);
+                    Element element = document.GetElement(id);
+                    if (element is FamilyInstance)
+                    {
+                        FamilyInstance famInst = element as FamilyInstance;
+                        IEnumerable<ElementId> nestedElemIds = famInst.ElementIdsOfMemberElements();
+                        elemIds.AddRange(nestedElemIds);
+                    }
+                }
+                elementIds = elemIds;
```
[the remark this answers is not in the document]
enarhi: changed

Revit_Core_Engine/Query/OpeningSurface.cs
enarhi: The usage of the first host instance here is just for placing the dummy instance. The later usage of all hosts is to extract the total opening area resulting from the dummy instances cutout on all hosts in case it overlaps multiple hosts.

Revit_Core_Engine/Query/OpeningSurface.cs
enarhi: I needed a way to effectively copy parameters, and as far as I could tell this is the best available method so that we can take advantage of the SetParameters method to manage setting of different parameter types. The hardcoded exclusion of Sill Height and Head Height is due to these causing errors if set through the api; they do not need setting anyhow as setting location and the other params places the instance in the exact location without touching these two parameters. If we have a better means of effectively copying revit params from one element and assigning to another I'm all ears!

Revit_Core_Engine/Query/OpeningSurface.cs
enarhi: done

enarhi:
@pawelbaran changes made, should be ready for re-review. I'm adding a revit test file as you noted to the pr as well, this will contain only pile caps with piles and nested windows families for now but more to come. One of the nested windows in the test file will not pull nested family geometry through as it currently stands. This is intentionally including to indicate current limitations, as the family that will not pull geometry is not a hosted window and therefore does not have a wall it belongs to or cuts. This would be similar to the equivalent limitation on a non-nested window family that is non-hosted, and should be covered in the future by a separate issue not specific to nested families.

Revit_Core_Engine/Query/OpeningSurface.cs
pawelbaran: I believe you could simply use in-built Revit get/set parameter functionality - no need to convert to BHoM parameters, it is just an overhead as far as I understand. If you need to skip any parameters, I recommend using [BuiltInParameter enum](https://www.revitapidocs.com/2020/fb011c91-be7e-f737-28c7-3f1e1917a0e0.htm) instead of string.

Revit_oM/Parameters/RevitIdentifiers.cs
pawelbaran: Hmm, I agree that the name fits into the convention, I am just a bit afraid of ambiguity. But I will not argue any more as I do not have any reasonable alternatives atm - the only thing I would recommend is to highlight that this property will only be filled for nested families (_owns_ suggests the concept of ownership, which may be understood in different ways).

pawelbaran:
I looked at the test file - the functionality makes sense, so my only concerns are with the code itself at the moment. Looking forward to getting the comments resolved and more testing being done by the users (you being a prime one @enarhi 😉).

A forward-reaching thought: it would be nice to have a method that would somehow group the BHoMObjects based on their nesting chain, so that the user could easily find all children and grandchildren of a given parent or simply group all objects that have a common root. How does that sound?

Revit_oM/Parameters/RevitIdentifiers.cs
enarhi: Agreed, description updated.

Revit_Core_Engine/Query/OpeningSurface.cs
enarhi: I've removed the redundant to and from bhom parameter portion. This required boilerplating the switches for the different Revit parameter value types in order to get the param values to then set them. Perhaps worth creating a method that handles that for us? Revit API of course does not provide for something as simple as SetParameter(element, parameter) which would streamline the code here.

enarhi: BuiltInEnum is now used for the Head and Sill Height exclusion as well, good call there!

pawelbaran (changes requested):
A few final comments left, I will also double-test the functionality with the routines that I have used before. Besides that, looks like we are getting there! 👍

pawelbaran (approved):
Tested, works as expected - LGTM now, both code and functionality wise 👍 Happy to merge as soon as we get a second review from BEnv/MEP (@FraserGreenroyd @michaelhoehn) + Versioning_Toolkit PR is ready as mentioned in [one of my previous comments](https://github.com/BHoM/Revit_Toolkit/pull/906/files#r525076244).

enarhi:
Versioning is ready as referenced above

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd (approved):
LGTM

PR #913, opened 2020-11-23
Included MEPCurve ConnectionProperty Information

### LINKED ISSUE #911: Include connector information in MEPCurve convert
#### Description:
When the prototype convert for `MEPCurve` was created a few weeks ago the connector information was left out for simplicity purposes. Now is the time to tackle the issue and include it properly.

tiagogrossi:

Closes #911 

Adds the proper information to the `ConnectionProperty` to the `MEPCurve` converts. At the moment only `CableTray` is using the feature due to the lack of the property in `Duct` and `Pipe` objects.
Only the properties `IsStartConnected` and `IsEndConnected` have been assigned, the properties about the connection point have been ignored as they are irrelevant in this object and overall redundant.

### Test files
Link [here](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Revit_Toolkit/Revit_Toolkit-Issue911-IncludedMEPCurveConnectorInformation?csf=1&web=1&e=5Bjkit). Check is visual, in the GH script I've separated a few cases to check and the values expected.

### Changelog
Modified `BH.Revit.Engine.Core.Convert.CableTrayFromRevit`
Modified `BH.Revit.Engine.Core.Convert.DuctFromRevit`
Modified `BH.Revit.Engine.Core.Convert.PipeFromRevit`
Modified `BH.Revit.Engine.Core.Query.LocationCurve`

### DIFF
--- a/Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
+++ b/Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
@@ -20,6 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.ComponentModel;
@@ -32,6 +33,7 @@ using BH.oM.MEP.System.ConnectionProperties;
 using BH.oM.MEP.System;
 using BH.oM.Reflection.Attributes;
 using BH.oM.Geometry;
+
 
 namespace BH.Revit.Engine.Core
 {
@@ -70,14 +72,14 @@ namespace BH.Revit.Engine.Core
             bool isStartConnected = false;
             bool isEndConnected = false;
             List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitCableTray, out isStartConnected, out isEndConnected, settings);
-
+            Vector revitCableTrayVector = BH.Engine.Geometry.Modify.RoundCoordinates(
+                VectorFromRevit((revitCableTray.Location as LocationCurve).Curve.GetEndPoint(0) -
+                                (revitCableTray.Location as LocationCurve).Curve.GetEndPoint(1)),4).Normalise();
             if (queried.Count > 2)
             {
                 //required to assert connector property later
                 queried = MatchRevitOrder(queried, revitCableTray);
             }
-            
-            Vector revitCableTrayVector = ((revitCableTray.Location as LocationCurve).Curve as Autodesk.Revit.DB.Line).Direction.VectorFromRevit().RoundCoordinates(4);
             
             for (int i = 0; i < queried.Count; i++)
             {
@@ -90,14 +92,13 @@ namespace BH.Revit.Engine.Core
                     ConnectionProperty = new CableTrayConnectionProperty(),
                     OrientationAngle = orientationAngle
                 };
-                
                 Vector bhomCableTrayVector = BH.Engine.Geometry.Modify.RoundCoordinates((thisSegment.StartPoint - thisSegment.EndPoint),4).Normalise();
 
                 if (queried.Count > 1)
                 {
-                    if (i == 0) //meaning it's the start segment of the revit cable tray that was split
+                    if (i == 0)
                     {
-                        if (BH.Engine.Geometry.Query.IsEqual(revitCableTrayVector, bhomCableTrayVector))
+                        if (revitCableTrayVector == bhomCableTrayVector)
                         {
                             thisSegment.ConnectionProperty.IsStartConnected = isStartConnected;
                             thisSegment.ConnectionProperty.IsEndConnected = true;
@@ -108,9 +109,9 @@ namespace BH.Revit.Engine.Core
                             thisSegment.ConnectionProperty.IsEndConnected = isStartConnected;   
                         }
                     }
-                    else if (i == queried.Count - 1) //meaning it's the end segment of the revit cable tray that was split
+                    else if (i == queried.Count - 1)
                     {
-                        if (BH.Engine.Geometry.Query.IsEqual(revitCableTrayVector, bhomCableTrayVector))
+                        if (revitCableTrayVector == bhomCableTrayVector)
                         {
                             thisSegment.ConnectionProperty.IsStartConnected = true;
                             thisSegment.ConnectionProperty.IsEndConnected = isEndConnected;
@@ -121,7 +122,7 @@ namespace BH.Revit.Engine.Core
                             thisSegment.ConnectionProperty.IsEndConnected = true;   
                         }
                     }
-                    else //meaning it's all mid segments of the revit cable tray that was split
+                    else
                     {
                         thisSegment.ConnectionProperty.IsStartConnected = true;
                         thisSegment.ConnectionProperty.IsEndConnected = true;
@@ -129,7 +130,7 @@ namespace BH.Revit.Engine.Core
                 }
                 else
                 {
-                    if (BH.Engine.Geometry.Query.IsEqual(revitCableTrayVector, bhomCableTrayVector))
+                    if (revitCableTrayVector == bhomCableTrayVector)
                     {
                         thisSegment.ConnectionProperty.IsStartConnected = isStartConnected;
                         thisSegment.ConnectionProperty.IsEndConnected = isEndConnected;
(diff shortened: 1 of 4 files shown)


Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -69,7 +69,17 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
             // Orientation angle
             double orientationAngle = revitCableTray.OrientationAngle(settings);
 
-            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitCableTray, settings);
+            bool isStartConnected = false;
+            bool isEndConnected = false;
+            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitCableTray, out isStartConnected, out isEndConnected, settings);
+            Vector revitCableTrayVector = BH.Engine.Geometry.Modify.RoundCoordinates(
+                VectorFromRevit((revitCableTray.Location as LocationCurve).Curve.GetEndPoint(0) -
+                                (revitCableTray.Location as LocationCurve).Curve.GetEndPoint(1)),4).Normalise();
```
pawelbaran, line 77: Could we please avoid splitting a single statement into 3 lines? 

I would also consider rewriting the query to make the code a bit more readable:
```suggestion
            Vector revitCableTrayVector = ((revitCableTray.Location as LocationCurve)?.Curve as Line)?.Direction.VectorFromRevit().RoundCoordinates(4);
```

On top of that, this variable is not used until line 84, so could be declared later.

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -79,9 +89,58 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
                     StartPoint = segment.StartPoint(),
                     EndPoint = segment.EndPoint(),
                     SectionProperty = sectionProperty,
-                    ConnectionProperty = null,
+                    ConnectionProperty = new CableTrayConnectionProperty(),
                     OrientationAngle = orientationAngle
                 };
+                Vector bhomCableTrayVector = BH.Engine.Geometry.Modify.RoundCoordinates((thisSegment.StartPoint - thisSegment.EndPoint),4).Normalise();
```
pawelbaran, line 95: I would consider migrating this whole part into a separate method, particularly in the case it could possibly be reused for other MEP types.

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -79,9 +89,58 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
...
+
+                if (queried.Count > 1)
+                {
+                    if (i == 0)
+                    {
+                        if (revitCableTrayVector == bhomCableTrayVector)
```
pawelbaran, line 101: This equality query is dangerous. I would rather use `BH.Engine.Geometry.Query.IsEqual` with tolerance.

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -94,6 +153,22 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
...
+        /***************************************************/
+        /****              Private Methods              ****/
+        /***************************************************/
+        
+        [Description("Re-orders converted BHoM Line list to match the direction from the reference MEPCurve.")]
+        private static List<BH.oM.Geometry.Line> MatchRevitOrder(List<BH.oM.Geometry.Line> linesToMatch, MEPCurve reference)
+        {
+            LocationCurve locationCurve = reference.Location as LocationCurve;
+            Curve curve = locationCurve.Curve;
+            BH.oM.Geometry.Point referenceStart = BH.Revit.Engine.Core.Convert.PointFromRevit(curve.GetEndPoint(0));
```
pawelbaran, line 165: You can make use of the extension methods to make the code more readable:
```suggestion
            BH.oM.Geometry.Point referenceStart = curve.GetEndPoint(0).PointFromRevit();
```

Revit_Core_Engine/Convert/Environment/FromRevit/Duct.cs
```diff
@@ -58,7 +58,9 @@ public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.M
                 bhomDucts = new List<BH.oM.MEP.System.Duct>();
             }
 
-            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitDuct, settings);
+            bool isStartConnected = false;
+            bool isEndConnected = false;
+            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitDuct, out isStartConnected, out isEndConnected, settings);
```
pawelbaran, line 63: Will `isStartConnected` and `isEndConnected` be relevant for this convert at any point? If not, I would consider adding an overload without `out` parameters that would call the one with them and simply not return them.

This comment refers to other converts as well.

pawelbaran (changes requested):
I went through the code as a first round of review - please see the code comments. Would be good to have some more discipline-aware reviewers involved too, to make sure that I am not derailing the concept 🙈 

@FraserGreenroyd @kayleighhoude

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
tiagogrossi: You're quite right, but I'd prefer not act on this if you're ok with it. Reason being that the other MEP obj don't take advantage of connector information. Also the only reason I'm doing this is to be more efficient in the path finding, since otherwise I'd have to query this info during it's runtime, so it saves time to grab it during the converts.

pawelbaran: Makes sense after the 2nd look at the code. Happy to keep it as is then, can be migrated out to a separate method in case of other converts wanting to use it too.

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -94,6 +152,20 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
             return bhomCableTrays;
         }
 
+        /***************************************************/
+        /****              Private Methods              ****/
+        /***************************************************/
+        
+        [Description("Re-orders converted BHoM Line list to match the direction from the reference MEPCurve.")]
+        private static List<BH.oM.Geometry.Line> MatchRevitOrder(List<BH.oM.Geometry.Line> linesToMatch, MEPCurve reference)
```
pawelbaran, line 160: Could you please embed this method inside `LocationCurveMEP`? I believe this should be a standard, not cabletray-specific.

Revit_Core_Engine/Convert/Environment/FromRevit/CableTray.cs
```diff
@@ -94,6 +152,20 @@ public static List<BH.oM.MEP.System.CableTray> CableTrayFromRevit(this Autodesk.
...
+        {
+            LocationCurve locationCurve = reference.Location as LocationCurve;
+            Curve curve = locationCurve.Curve;
+            BH.oM.Geometry.Point referenceStart = curve.GetEndPoint(0).PointFromRevit();
+
+            return linesToMatch.OrderBy(x => x.Start.Distance(referenceStart)).ToList();
```
pawelbaran, line 166: Is that a robust solution? What if the curves had some weird snail-like shape? Having said that, couldn't you joint the curves into a `PolyCurve`, flip it so that its start point is closer to the start point of element's location curve and then explode?

A bit of an overhead, but may be worth - please dismiss this comment if that does not make sense to you.

Revit_Core_Engine/Convert/Environment/FromRevit/Duct.cs
```diff
@@ -57,7 +57,7 @@ public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.M
             {
                 bhomDucts = new List<BH.oM.MEP.System.Duct>();
             }
-
+
```
pawelbaran, line 60: 👀

pawelbaran (changes requested):
The code makes sense to me now, I would  only consider elevating `MatchRevitOrder` a bit - please see the code comments. Besides that good to go, let's wait for the functional review!

tiagogrossi:
@pawelbaran thanks for the review. I've just commit the latest changes to address the comments. You'll notice that in the `CableTrayFromRevit` file, line 74, I've reversed to the manner I was retrieving the revit object vector from before (start-end.normalize), because for some reason `LocationCurve.Line.Direction` was retrieving some awkward results. Hope that's ok for you.

pawelbaran (commented):
LGTM code wise 👍 However, I will keep this review to avoid getting the PR merged before the functionality gets approved by the discipline (@FraserGreenroyd @michaelhoehn @kayleighhoude).

michaelhoehn (approved):
Happy to approve after functionality tests with the provided scripts. Connection properties seem to correspond correctly 😄

pawelbaran (approved):

Issue #992, opened 2021-03-17
Revit_Adapter: Support UniformLoadSet mapping to Filled Region Type


JosefTaylor:
#### Description:

Once `UniformLoadSet` is promoted per [BHoM Issue 1224](https://github.com/BHoM/BHoM/issues/1224), I would like to be able to push this object to Revit and have it mapped to a filled region type. This would need to create a new filled region type and create project parameters match the names of the loads in the `UniformLoadSet`

pawelbaran:
As discussed with @JosefTaylor and mentioned in [this comment](https://github.com/BHoM/Revit_Toolkit/issues/993#issuecomment-814166106) the idea is to enable creation of Revit types in general, or at least creation of `FilledRegionTypes` in general - in such case, resolving this issue will require creating a method that will translate the `UniformLoadSet` properties into a general filled region type to be pushed to Revit, all happening on the BHoM/GH side, and not on convert stage.

TBC once more investigation is made around #993.

pawelbaran:
@JosefTaylor it this something you still find useful? Or would you rather cover the workflow using the existing GH and BHoM components?

PR #1027, opened 2021-05-12
Parameter creation and cloning of element types enabled

### LINKED ISSUE #993: Revit_Adapter: Support creation of Filled Region Types and creation of new project parameters
#### Description:

In order to push load map data from Excel (where it is authored) to Revit, I need to create new filled region types corresponding to each set of loads, and create type parameters for Detail Items (unfortunately I cannot contain them to Filled Region Types). Once these are created, I can update them from BHoM with the correct hatch, type mark, parameter values, etc., but there's a lot of effort in creating these (and getting all the names right).

Generalized, this means creating new `InstanceProperties` i.e. Family Types in Revit and also creating new project parameters.

Alternately, a solution specific to Filled Regions is possible: a new BHoM object `FilledRegionType` which can be pushed to Revit. This still requires the general ability to add project parameters

Lastly, another approach would be to create a new Revit family more suited to load map use, which would have parameters built in with generic names (e.g. LoadArea Value1, LoadArea Value2, etc.), which could be ether renamed to match the names of loads, or could be scheduled in a way as to change the names of headers to names stored in another set of parameters (e.g. LoadArea Name1, LoadArea Name2, etc.)

pawelbaran:
### NOTE: Depends on 

Not a direct dependency, but testing the colour parameter assignment requires building https://github.com/BHoM/BHoM_Engine/pull/2502.

   
Closes #993

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FRevit%5FToolkit%2F%231027%2DProjectParametersAndRegionTypes&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) - best to test with the Autodesk basic sample architecture project (_C:\Program Files\Autodesk\Revit 20XX\Samples\rac_basic_sample_project.rvt_). Please note that you may need to switch the shared parameters file to a new/empty one because the original is blocked (Revit_Toolkit will inform you about it).

Please feel free to do more stress testing and add more test files to the folder.

### Changelog

- cloning of types enable using `ClonedType` object
- assignment of colour parameters enabled
- pushing of Revit parameter definitions (project and shared parameters) enabled based on `ParameterDefinition` object

### Additional comments

@JosefTaylor it would be great if you could review this PR in the context of your needs expressed in #992 - this PR should be a basis on which you should be able to create methods that will extract relevant information from `UniformLoadSet` and prepare it for pushing to Revit.

### DIFF
--- a/Revit_Core_Engine/Convert/Revit/ToRevit/ElementType.cs
+++ b/Revit_Core_Engine/Convert/Revit/ToRevit/ElementType.cs
@@ -22,6 +22,7 @@
 
 using Autodesk.Revit.DB;
 using BH.Engine.Adapters.Revit;
+using BH.oM.Adapters.Revit;
 using BH.oM.Adapters.Revit.Properties;
 using BH.oM.Adapters.Revit.Settings;
 using System;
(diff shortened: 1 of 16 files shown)


pawelbaran:
@BHoMBot check required

JosefTaylor:
Basic mechanics work great! A couple difficulties I'm running into:

1. I can't seem to set multiple parameters by providing a single object, a list of parameter names, and a list of parameter values; this is going to be very important for this workflow. Is there something I'm missing in that component?
2. Pushing a cloned filled region type fails if the type already exists- is there a way to update? Maybe the solution is to create the clones, pull the resulting filledregiontypes and then update their parameters.

pawelbaran:
Thanks for looking into @JosefTaylor, please see my responses below.

> 1. I can't seem to set multiple parameters by providing a single object, a list of parameter names, and a list of parameter values; this is going to be very important for this workflow. Is there something I'm missing in that component?

Do you mean `SetRevitParameter`? If so, then you could use its sister method `SetRevitParameters` for multiple inputs.

> 2. Pushing a cloned filled region type fails if the type already exists- is there a way to update? Maybe the solution is to create the clones, pull the resulting filledregiontypes and then update their parameters.

Yeah, you are right that you could pull the filled region type and update it, alternatively, you can use the output of Push:
image

On top of that, yesterday I forgot to push one commit plus I am going push one more, so please pull latest before further testing 🙈

pawelbaran:
@BHoMBot check required

JosefTaylor (approved):
Ran out of time to compile the latest two commits but was happy with functionality before then. If someone can check that they don’t break anything I’m happy!

pawelbaran:
Further commits were minor bug fixes, so I am quite comfortable without further in-depth review. However, will keep the PR open until double checking with @JosefTaylor that we are on the same page with that, as well as with the next steps needed to resolve #992.

pawelbaran:

pawelbaran:

JosefTaylor (approved):

pawelbaran:
@JosefTaylor can we assume the PR was reviewed and tested, so that I could merge it?

Issue #1012, opened 2021-04-06
Introduce converts for BH.oM.Architecture.BuildersWork.Opening


pawelbaran:
#### Description:

As in the title - to be figured out how to do this most smartly.

PR #1055, opened 2021-07-23
BHoM type vs Revit family name mapping implemented, ParameterSettings renamed + BuildersWork.Opening convert added

### LINKED ISSUE #1054: Refactor the parameter mapping system and conversion from IRequest to ElementIds on Pull to allow matching types by parameters
#### Description:

As discussed many times, in case of some Revit element types, it is hard to map them 1:1 with BHoM objects. A flag example for that is `FamilyInstance`, which can be virtually anything. Some hint can be given by the category to which it belongs (i.e. _Structural Framing_ is quite obvious), but in some cases this is of no help either, e.g. _Generic Component_, which can be virtually anything or _Security Device_, encompassing multiple subtypes of objects.

A solution to that is setting up a framework where certain BHoM types can be linked with chosen Revit families, followed by mapping of the BHoM object properties against the parameters of these families. Today we had a good chat on that topic with @tiagogrossi and I think I have a solution to that. Stay tuned!

@kayleighhoude @michaelhoehn

pawelbaran:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1258
   
Closes #1054
Half of #1012

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FRevit%5FToolkit%2F%231055%2DTypeAndParameterMappingAlt&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### Changelog

- BHoM type vs Revit family name mapping implemented
- `ParameterSettings` renamed to `MappingSettings`
- `FromRevit` convert added for `Architecture.BuildersWork.Opening`

### Additional comments

A few points on the mechanics to clarify the proposed solution:
- all parameter and type mapping is stored in `MappingSettings`, which is a property of the adapter
- filtering of the elements to pull is sole responsibility of the `IRequest` which is not shared with adapter settings at the moment, therefore mapping does not apply to filtering of the elements (and is not likely to apply in the future)
- Revit family and parameter mapping is separated because:
    - family mapping is only relevant on `Pull` of the elements that cannot be clearly linked to any BHoM type without this extra hint
    - parameter mapping applies both on `Pull` and `Push`, and is an entirely separate mechanism

Not ready to merge yet, but would be great to start checking the functionality @tiagogrossi


tiagogrossi:
I've had an initial run through this and found some issues in the sample file, I'll write things down here a bit dryly, happy expand on a call. 
- Both `ParameterMap` and `FamilyMap` showed error "The component is not linked to a method". Am I missing some build?
- In `RevitSettings` the input name still says `parameterSettings` instead of `mappingSettings`
- In the first pull the sample uses as request a `FilterByFamilyAndTypeName`. This is confusing because I was expecting simply the type opening as input. If we had a bunch of instances of Generic Model and we simply input the opening type, what would happen? I'd test but the whole thing is not working in my computer it seems.

pawelbaran:
@tiagogrossi it looks like the code did rebuild correctly on your machine? I have just double checked on my end and the file is OK.

Regarding the request - this is a tricky case. As mentioned in the PR description:
 > filtering of the elements to pull is sole responsibility of the `IRequest` which is not shared with adapter settings at the moment, therefore mapping does not apply to filtering of the elements (and is not likely to apply in the future)

...which in practice means that the request for openings would not know about the adapter settings, therefore would not map the families. Theoretically we could intertwine the filtering process and the adapter settings, which would resolve this particular issue, but this would only work for family mapping, not the parameters, i.e. `FilterByParameterNumber` would never take parameter mapping into account because at this stage we do not know which map to apply as it is dependent on the BHoM type we convert to, which is not known at that stage. I see this really confusing: _why are the families mapped on filtering while the parameters are not?_ And not sure how to resolve it, taken the we never know the output type in advance.

Does the above make sense? If so, maybe you see a solution to it from your wider (usually better 😉) perspective?

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check versioning

pawelbaran:
I am closing this PR and killing the branch because it fell out of sync after #1058 got merged + https://github.com/BHoM/BHoM/pull/1258 and https://github.com/BHoM/BHoM_Engine/pull/2572 will be easier to merge without this dragging them back. This branch will be reworked and opened as a separate PR.

PR #1082, opened 2021-09-15
Add tolerance parameter for pull with PlanarSurface and IsClosed methods

### LINKED ISSUE #1008: Area calculation failing for some elements pulled from Revit due to open curve error when curve is actually closed
#### Description:
As per the title and the image below:
image

#### Steps to reproduce:
Certain elements pulled from Revit fail for area calculations due to perceived open curves that are actually closed. This appears to be specific to the method used for PlanarSurfaces. It goes away if the location is updated per a recreated surface from the edges derived from the failing surface as shown in the image above. On further inspection, this is due to tolerances and a gap between points of ~0.0004 per the Revit pull. Not sure the best fix for this, but it seems like one way or another the geometry pulled from revit can have small tolerance issues.

#### Expected behaviour:
Accurate area calculations on these elements as long as they are in fact defined by closed curves.

#### Test file(s):

[FailingAreaCalc.zip](https://github.com/BHoM/BHoM_Engine/files/6240069/FailingAreaCalc.zip)

### LINKED ISSUE #1062: Add tolerance usage for Create Planar Surface method
#### Description:
Once tolerance is added to the central create planar surface method, we should pass the tolerance from revit settings to that method to avoid instances where planar surface creation checks fail due to gaps in "closed" curves between our default tolerance and Revit's default tolerance.

enarhi:

Closes #1062 
Closes #1008 
(#1008 is closed primarily by the previous Tolerance parameter capabilities, as the issue can be solved with current Revit_Toolkit capabilities along with a tolerance input value on the IArea method that matches the RevitSettings.DistanceTolerance value accordingly.)

Allows for tolerance set in RevitSettings to be applied to Planar Surface creation, thereby allowing pulling closed regions such as a FilledRegion where the region is closed per Revit tolerances, but not per default BHoM tolerances.

### Test files
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/%231062-PlanarSurfaceTolerance?csf=1&web=1&e=kDTc7h

Test by checking with the different tolerances to confirm pull failure when tolerance value is too low, and success when tolerance is higher. Note that Rhino unit tolerance will drive the Surface component at the end, so you should set your tolerance low in Rhino (recommend 0.0001m) to ensure proper functionality (Rhino tolerance is inverse to the input revit tolerances here, since it drives the joining of curves to create the surface and a value too high causes errant Grasshopper-side snapping of curves when joining/converting to a surface.

### DIFF
--- a/Revit_Core_Engine/Convert/Revit/FromRevit/DraftingInstance.cs
+++ b/Revit_Core_Engine/Convert/Revit/FromRevit/DraftingInstance.cs
@@ -65,7 +65,7 @@ namespace BH.Revit.Engine.Core
                 curves.Add(loop.FromRevit());
             }
 
-            List<PlanarSurface> surfaces = BH.Engine.Geometry.Create.PlanarSurface(curves);
+            List<PlanarSurface> surfaces = BH.Engine.Geometry.Create.PlanarSurface(curves, settings.DistanceTolerance);
             if (surfaces.Count == 1)
                 draftingInstance = new DraftingInstance { Properties = instanceProperties, ViewName = view.Name, Location = surfaces[0] };
             else
--- a/Revit_Core_Engine/Convert/Revit/FromRevit/DraftingInstance.cs
+++ b/Revit_Core_Engine/Convert/Revit/FromRevit/DraftingInstance.cs
@@ -28,6 +28,7 @@ using BH.oM.Adapters.Revit.Settings;
 using BH.oM.Base;
 using BH.oM.Geometry;
 using BH.oM.Reflection.Attributes;
+using BH.Engine.Reflection;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


enarhi:
@BHoMBot check compliance
@BHoMBot check versioning
@BHoMBot check null-handling

enarhi:
@BHoMBot check compliance

enarhi:
@BHoMBot check required

Revit_Core_Engine/Convert/Revit/FromRevit/DraftingInstance.cs
```diff
@@ -65,7 +65,7 @@ public static DraftingInstance DraftingInstanceFromRevit(this FilledRegion fille
                 curves.Add(loop.FromRevit());
             }
 
-            List<PlanarSurface> surfaces = BH.Engine.Geometry.Create.PlanarSurface(curves);
+            List<PlanarSurface> surfaces = BH.Engine.Geometry.Create.PlanarSurface(curves, settings.DistanceTolerance);
```
pawelbaran, line 68: Before line 69,  could you please add a check if (surfaces.Count == 0) and if so, then raise a warning and set the instance's location to `null`? It seems to make more sense than returning the location equal to an empty `PolySurface`. Or am I confusing things?

pawelbaran (changes requested):
As usual, the issues related to tolerances and chains of messages turn to be trickier than they are! 🙈 I have found a few minor things that could be tweaked in your code (please see the comments), plus raised https://github.com/BHoM/BHoM_Engine/pull/2622 to avoid meaningless warnings to show up (see your example with too strict tolerance).

On top of that, could you please be so kind and provide a test case for `AnalyticalOutlines` query? Thanks!

enarhi:
Added all changes per your comments @pawelbaran , please rereview.

enarhi:
@BHoMBot check compliance

enarhi:
@BHoMBot check versioning
@BHoMBot check null-handling

enarhi:
@pawelbaran the test files have been updated to include the case for AnalyticalOutlines, which previously failed when pulling things such as floors where the small gap issue occurs. It seems to happen most often for consecutive joining arcs along a CurveLoop, for all that's worth.

pawelbaran (approved):
All LGTM now 👍

pawelbaran:
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check serialisation
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

PR #1088, opened 2021-09-15
Create.RevitPullConfig versioned

### LINKED ISSUE #1087: Create.RevitPullConfig has not been versioned
#### Description:

`BH.Engine.Adapters.Revit.Create.RevitPullConfig` has changed a few times over the time - currently, in 4.3, support for versions 3.3 and 4.0 is over, so the components from those versions appear as magic components. Although this is according to our versioning strategy, this method is present in almost any script, so would be great to make it not break after the update. Actually, this is deadly simple and can be done using `PreviousVersionAttribute`, so I believe it is worth giving it a go.

pawelbaran:

Closes #1087

### Test files

The origin of the issue is _PullFramingProfiles.gh_ from [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F03%5FAlpha%2FBHoM%2FRevit%5FToolkit%2FPull&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

Test of another version has been added [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FRevit%5FToolkit%2F%231088%2DCreateRevitPullConfigVersioning&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4). Since the input order changed between the versions, I am adding a screenshot from the old version too:
image

### DIFF
--- a/Revit_Engine/Create/Config/RevitPullConfig.cs
+++ b/Revit_Engine/Create/Config/RevitPullConfig.cs
@@ -32,7 +32,9 @@ namespace BH.Engine.Adapters.Revit
         /***************************************************/
         /****              Public methods               ****/
         /***************************************************/
-        
+
+        [PreviousVersion("4.3", "BH.Engine.Adapters.Revit.Create.RevitPullConfig(BH.oM.Adapters.Revit.Enums.Discipline, System.Boolean, BH.oM.Adapters.Revit.PullGeometryConfig, BH.oM.Adapters.Revit.PullRepresentationConfig, System.Boolean)")]
+        [PreviousVersion("4.3", "BH.Engine.Adapters.Revit.Create.RevitPullConfig(BH.oM.Adapters.Revit.Enums.Discipline, System.Boolean, BH.oM.Adapters.Revit.PullGeometryConfig, BH.oM.Adapters.Revit.PullRepresentationConfig)")]
         [Description("Creates a pull action-specific configuration used for adapter interaction with Revit.")]
         [InputFromProperty("discipline")]
         [InputFromProperty("includeClosedWorksets")]


pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check null-handling
@BHoMBot check serialisation
@BHoMBot check versioning

Revit_Engine/Create/Config/RevitPullConfig.cs
```diff
@@ -32,7 +32,9 @@ public static partial class Create
         /***************************************************/
         /****              Public methods               ****/
         /***************************************************/
-        
+
+        [PreviousVersion("4.3", "BH.Engine.Adapters.Revit.Create.RevitPullConfig(BH.oM.Adapters.Revit.Enums.Discipline, System.Boolean, BH.oM.Adapters.Revit.PullGeometryConfig, BH.oM.Adapters.Revit.PullRepresentationConfig, System.Boolean)")]
+        [PreviousVersion("4.3", "BH.Engine.Adapters.Revit.Create.RevitPullConfig(BH.oM.Adapters.Revit.Enums.Discipline, System.Boolean, BH.oM.Adapters.Revit.PullGeometryConfig, BH.oM.Adapters.Revit.PullRepresentationConfig)")]
```
michaelhoehn, line 37: Do you need both of these attributes? Sorry, I'm far from a Versioning expert, but this isn't something I've seen.

FraserGreenroyd: I think the purpose of versioning is to try and fix for multiple versions going backwards. If prior to this version both were used and it now wants to upgrade from 3.3 to 4.3 and 4.0 to 4.3 in the same manner, then both attributes would be acceptable.

We've used it in the Engine a couple of times when moving a method from Environment_Engine and Structure_Engine into a common engine such as Analytical_Engine (to work on the generic panel), we have then used `PreviousVersion` to indicate both previous engine methods can now version into this new method.

Obviously that's taking 2 methods into 1 from the same version, this is similar only in that it is taking 2 methods from different versions into 1 for the current version. Based on my understanding of the issue at least. So could be wrong with what's needed, but certainly from a Versioning perspective, 2 attributes like this would be valid.

FraserGreenroyd (approved):
Check Versioning is happy so I'm happy.

michaelhoehn:
@BHoMBot check ready-to-merge

Revit_Engine/Create/Config/RevitPullConfig.cs
pawelbaran: Yeah this is exactly what is happening @FraserGreenroyd, 2 previous versions of the method not versioned but simply deprecated and removed, now turning red or magic on the canvas. Looks like we teamed up quite well: I knew what is needed, you confirmed that my 'give it a shot' solution makes sense 😃 Thanks both!

PR #1113, opened 2021-10-18
Revit_Engine: RevitDiffing to consider non-Revit objects and base the diff on RevitIdentifier

### LINKED ISSUE #1110: Revit_Engine: RevitDiffing to consider non-Revit objects and base the diff on RevitIdentifier
### Description

As discussed with @pawelbaran , the current Revit-specific Diffing implementation only considers Revit-specific objects (e.g. `ModelInstance`s).

However, I did not account for the fact that Revit is actually able to pull back non-Revit objects (e.g. `BH.oM.Physical.Wall`s) that come with a `RevitIdentifier` fragment attached.

This poses a challenge, for 2 reasons:
- Currently the base `BH.Engine.Diffing.IDiffing()`, when determining if an object should be diffed with a Toolkit-specific object, is only looking for the namespace of the top-level object. So for a `Wall` it would only look for a specific `BH.oM.Physical.Diffing()` method.
- The Revit-specific `Diffing` method, on the other hand, asks explicitly for the input objects to be Revit-specific (i.e. coming from `BH.oM.Adapters.*` namespace), so it would not work for `Walls`.

### Proposed changes
We can change the behaviour of the Revit-specific Diffing to remove the check on the objects being Revit-specific, and only ask for them to host a RevitIdentifier parameter.

By doing that, however, the base `BH.Engine.Diffing.IDiffing()` would still not be calling the Revit-specific Diffing. We discussed what could be done:
- an option could be that the base IDiffing looks for all the namespaces of all the properties and subproperties, and based on that does some dispatch to the namespace-specific diffing. However, this would be near-impossible to manage: what if a `Wall` has multiple properties coming from multiple namespaces? Also, consider the slowdown of this.
- Instead, we can rely on the presence of `IPersistentAdapterId` identifiers on the objects. We can get all the `IPersistentAdapterId`s stored on the objects, and group them by parent namespace. Then, because we also gather all the Toolkit-specific Diffing methods, we can check if for every namespace of `IPersistentAdapterId` there is a corresponding Toolkit-specific Diffing method. If so:
   -  If there is only one type of `IPersistentAdapterId` and a corresponding Toolkit-specific method was found, we can call that.
   - If there are multiple `IPersistentAdapterId` stored:
      - If there are no corresponding Toolkit-specific Diffing methods, just continue with the base IDiffing;
      - If there is only one corresponding Toolkit-specific Diffing, call that method after Noting that to the user.
      - If there are multiple corresponding Toolkit-specific Diffing methods, return a Warning to the User: 
        > _Multiple `PersistentAdapterId`s and corresponding Toolkit-specific methods were found. Please pick a Toolkit-specific method to use with these objects. IDiffing will continue with the automatic process._ 
      
          and continue with the IDiffing.

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM_Engine/pull/2656
   
Closes #1110 

### Test files

https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FRevit%5FToolkit%2F%231110%2DRevitDiffingOfNonRevitObjects&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4

### DIFF
--- a/Revit_Engine/Compute/RevitDiffing.cs
+++ b/Revit_Engine/Compute/RevitDiffing.cs
@@ -114,11 +114,42 @@ namespace BH.Engine.Adapters.Revit
                 return null;
             }
 
+            // Get only the BHoMObjects from the input collections.
+            IEnumerable<IBHoMObject> pastBHoMObjects = pastObjects.OfType<IBHoMObject>();
+            IEnumerable<IBHoMObject> followingBHoMObjects = followingObjects.OfType<IBHoMObject>();
+
+            // Make sure all input objects are BHoMObjects.
+            if (pastBHoMObjects.Count() != pastObjects.Count())
+            {
+                BH.Engine.Reflection.Compute.RecordError($"Some of the input {nameof(pastObjects)} were not BHoMObjects. This method only works with BHoMObjects.");
+                return null;
+            }
+
+            if (followingBHoMObjects.Count() != followingObjects.Count())
+            {
+                BH.Engine.Reflection.Compute.RecordError($"Some of the input {nameof(followingObjects)} were not BHoMObjects. This method only works with BHoMObjects.");
+                return null;
+            }
+
+            // Make sure all input BHoMObjects have a RevitIdentifier Fragment attached.
+            foreach (var o in pastBHoMObjects)
+                if (o.GetAllFragments(typeof(RevitIdentifiers)).Count < 1)
+                {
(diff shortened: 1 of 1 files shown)


pawelbaran:

PR #1154, opened 2021-12-22
Element type query centralised and made more logical

### LINKED ISSUE #1146: SetType should pass relevant categories into IElementType query
#### Description:

For example in the snippet below:
https://github.com/BHoM/Revit_Toolkit/blob/a71d99fa93ab5fbad0c01ac14885e090c186ab91/Revit_Core_Engine/Modify/SetType.cs#L53-L55

...the call to `IElementType` in the last line should take a list of relevant categories. Same applies to other `SetType` methods.

Actually, shouldn't the collection of categories be equal to the category of the existing element, assuming that the element cannot change its category?

pawelbaran:

Closes #1146

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FRevit%5FToolkit%2F%231154%2DSetTypeCleanup&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

### Changelog

- Revit element type search centralised for creation and update of the elements based on the BHoM objects
- informative warnings added to explain to the user what is the source of the element type information for given convert (BHoM object name vs name of its defining property like `Construction`)
- `Query.ElementType` stopped from modifying the queried Revit element types

### Additional comments

The problem proved to be much more complex than originally stated #1146. There were 3 parallel challenges to deal with:
- element type query being different on creation and update of the Revit elements based on BHoM objects
- not enough feedback  on how what is happening on type update
- element type query being able to modify the queried type

First two were rather straightforward to capture and fix, just a matter of pulling the strings together. The 3rd one, however, was a bit more tricky. Until now, when `Query.ElementType` was called on physical `Wall`, `Beam` etc., `ToRevitElementType` method was called from within. This resulted in an attempt to convert and overwrite the parameters of the existing type (if found) instead of simply querying it. This, I believe, is fundamentally wrong, so this has been changed to `Query.ElementType` only querying the element type, but never modifying it.

The above is a breaking change, but I believe it is a step in the right direction, plus it is likely to be left unnoticed by the users. The only scenario affected by the change is a `Push` of an object with its defining property (e.g. `Construction`) having modified Revit parameter values (not properties) using `SetRevitParameter` method. Now, after the change, `Push` of elements affects the elements, while modification of element types can be achieved by pushing the element types themselves.

Sorry for the non-Revit users trying to understand the above 🙈

### DIFF
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -96,7 +96,7 @@ namespace BH.Revit.Engine.Core
 
             Line columnLine = framingElement.Location.IToRevit() as Line;
 
-            FamilySymbol familySymbol = framingElement.ElementType(document, settings, refObjects);
+            FamilySymbol familySymbol = framingElement.ElementType(document, settings);
             if (familySymbol == null)
             {
                 Compute.ElementTypeNotFoundWarning(framingElement);
@@ -200,7 +200,7 @@ namespace BH.Revit.Engine.Core
 
             Level level = document.LevelBelow(framingElement.Location, settings);
 
-            FamilySymbol familySymbol = framingElement.ElementType(document, settings, refObjects);
+            FamilySymbol familySymbol = framingElement.ElementType(document, settings);
             if (familySymbol == null)
             {
                 Compute.ElementTypeNotFoundWarning(framingElement);
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -96,10 +96,7 @@ namespace BH.Revit.Engine.Core
 
             Line columnLine = framingElement.Location.IToRevit() as Line;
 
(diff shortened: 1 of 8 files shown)


pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check serialisation
@BHoMBot check versioning

pawelbaran:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check installer
@BHoMBot check serialisation
@BHoMBot check versioning

michal-pekacki (dismissed):
Tested on different Revit versions and LGTM! 👍

pawelbaran:
@BHoMBot check required

michal-pekacki (approved):
LGTM 👍

pawelbaran:
@BHoMBot check ready-to-merge

PR #1190, opened 2022-04-11
Enable pushing columns flipped upside down

### LINKED ISSUE #1151: Enable pushing columns flipped upside down
#### Description:

Currently when the start point of a column is above its end point, an error is thrown. Instead, a warning could be thrown and the issue fixed.

vietle-bh:

Closes #1151

### Test files

PR-specific test files on [SharePoint](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/%231190-EnablePushingColumnsFlippedUpsideDown?csf=1&web=1&e=fI86kS).

### Changelog

- `FamilyInstance.cs` will auto-flip a column's location line if its start point is higher than its endpoint. 

### Additional comments
Flipped geometry is saved back to the original `BH.oM.Physical.Elements.Column` object, allowing subsequent Revit column generation to succeed. Does this need updating anywhere else?

### DIFF
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -93,7 +93,7 @@ namespace BH.Revit.Engine.Core
             else if (locationLine.Start.Z > locationLine.End.Z)
             {
                 BH.Engine.Base.Compute.RecordWarning(string.Format("Revit column's start point needs to be higher than its end points. Consider flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                framingElement.Location = locationLine.Flip();
+                locationLine = locationLine.Flip();
             }
 
             Level level = document.LevelBelow(framingElement.Location, settings);
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -92,15 +92,15 @@ namespace BH.Revit.Engine.Core
             }
             else if (locationLine.Start.Z > locationLine.End.Z)
             {
-                BH.Engine.Base.Compute.RecordNote(string.Format("The bottom of the input column was above its top. Its location line has been flipped to allow creating the Revit element. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                framingElement.Location = locationLine.Flip();
+                BH.Engine.Base.Compute.RecordNote(string.Format("The bottom of the input column was above its top. Its location line was flipped to allow creating the Revit element. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                locationLine = locationLine.Flip();
             }
 
-            Level level = document.LevelBelow(framingElement.Location, settings);
+            Level level = document.LevelBelow(locationLine, settings);
             if (level == null)
                 return null;
 
-            Line columnLine = framingElement.Location.IToRevit() as Line;
+            Line columnLine = locationLine.IToRevit() as Line;
 
             FamilySymbol familySymbol = framingElement.ElementType(document, settings);
             if (familySymbol == null)
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -72,18 +72,6 @@ namespace BH.Revit.Engine.Core
 
             settings = settings.DefaultIfNull();
 
-            if (framingElement.Location == null)
-            {
-                BH.Engine.Base.Compute.RecordError(String.Format("Revit element could not be created because the driving curve of a BHoM object is null. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                return null;
-            }
-
-            if (framingElement.Location as BH.oM.Geometry.Line == null)
-            {
-                BH.Engine.Base.Compute.RecordError(string.Format("Revit does only support line-based columns. Try pushing your element as a beam instead. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                return null;
-            }
-
             BH.oM.Geometry.Line locationLine = framingElement.ColumnLine();
             if (locationLine == null)
                 return null;
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -85,15 +85,11 @@ namespace BH.Revit.Engine.Core
             }
 
             BH.oM.Geometry.Line locationLine = (BH.oM.Geometry.Line)framingElement.Location;
-            if (locationLine.Start.Z == locationLine.End.Z)
+            locationLine = locationLine.SortColumEndpoints();
+            if (locationLine == null)
             {
-                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same elevation. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                BH.Engine.Base.Compute.RecordError(string.Format("Column line's start and end points have the same elevation. Conversion failed for BHoM_Guid: {0}", framingElement.BHoM_Guid));
                 return null;
-            }
-            else if (locationLine.Start.Z > locationLine.End.Z)
-            {
-                BH.Engine.Base.Compute.RecordNote(string.Format("The bottom of the input column was above its top. Its location line was flipped to allow creating the Revit element. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                locationLine = locationLine.Flip();
             }
 
             Level level = document.LevelBelow(locationLine, settings);
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -87,7 +87,7 @@ namespace BH.Revit.Engine.Core
             BH.oM.Geometry.Line locationLine = (BH.oM.Geometry.Line)framingElement.Location;
             if (locationLine.Start.Z == locationLine.End.Z)
             {
-                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same height. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same elevation. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
                 return null;
             }
             else if (locationLine.Start.Z > locationLine.End.Z)
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -84,10 +84,16 @@ namespace BH.Revit.Engine.Core
                 return null;
             }
 
-            if (((BH.oM.Geometry.Line)framingElement.Location).Start.Z >= ((BH.oM.Geometry.Line)framingElement.Location).End.Z)
+            BH.oM.Geometry.Line locationLine = (BH.oM.Geometry.Line)framingElement.Location;
+            if (locationLine.Start.Z == locationLine.End.Z)
             {
-                BH.Engine.Base.Compute.RecordError(string.Format("Start point of Revit columns need to have lower elevation than the end point. Have a look at flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same height. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
                 return null;
+            }
+            else if (locationLine.Start.Z > locationLine.End.Z)
+            {
+                BH.Engine.Base.Compute.RecordWarning(string.Format("Revit column's start point needs to be higher than its end points. Consider flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                framingElement.Location = locationLine.Flip();
             }
 
             Level level = document.LevelBelow(framingElement.Location, settings);
--- a/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
+++ b/Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
@@ -92,8 +92,8 @@ namespace BH.Revit.Engine.Core
             }
             else if (locationLine.Start.Z > locationLine.End.Z)
             {
-                BH.Engine.Base.Compute.RecordWarning(string.Format("Revit column's start point needs to be higher than its end points. Consider flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
-                locationLine = locationLine.Flip();
+                BH.Engine.Base.Compute.RecordNote(string.Format("The bottom of the input column was above its top. Its location line has been flipped to allow creating the Revit element. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                framingElement.Location = locationLine.Flip();
             }
 
             Level level = document.LevelBelow(framingElement.Location, settings);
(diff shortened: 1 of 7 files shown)


vietle-bh:
@BHoMBot check compliance

vietle-bh:
@BHoMBot check required

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -84,11 +84,17 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
                 return null;
             }
 
-            if (((BH.oM.Geometry.Line)framingElement.Location).Start.Z >= ((BH.oM.Geometry.Line)framingElement.Location).End.Z)
+            BH.oM.Geometry.Line locationLine = (BH.oM.Geometry.Line)framingElement.Location;
+            if (locationLine.Start.Z == locationLine.End.Z)
             {
-                BH.Engine.Base.Compute.RecordError(string.Format("Start point of Revit columns need to have lower elevation than the end point. Have a look at flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same height. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
```
pawelbaran, line 90:
```suggestion
                BH.Engine.Base.Compute.RecordError(string.Format("Column's start and end points have the same elevation. Conversion failed. BHoM_Guid: {0}", framingElement.BHoM_Guid));
```

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -84,11 +84,17 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
                 return null;
             }
+            else if (locationLine.Start.Z > locationLine.End.Z)
+            {
+                BH.Engine.Base.Compute.RecordWarning(string.Format("Revit column's start point needs to be higher than its end points. Consider flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
```
pawelbaran, line 95: This message does not make much sense: you raise the warning and then flip the curve anyways. I would rather use `RecordNote` instead of `RecordWarning` and simply inform the user that the bottom of the input column was above its top but this has been aligned when creating the Revit element.

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -84,11 +84,17 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
+                framingElement.Location = locationLine.Flip();
```
pawelbaran, line 96: This may be pretty dangerous as you are changing the input object - this should be avoided, so if you want this to do you should either clone the input BHoM object at the beginning (heavy, not recommended) or rather flip the location live itself and use it as an input for the Revit column further downstream:
```suggestion
                locationLine = locationLine.Flip();
```

pawelbaran (changes requested):
Looks like you identified the right place to intervene @vietle-bh 👍 However, I have a few comments on the changes you have made - please see my code comments.

Revit_Engine/Modify/SortEndpoints.cs
```diff
@@ -0,0 +1,58 @@
...
+                BH.Engine.Base.Compute.RecordError(string.Format("Line's start and end points have the same elevation.}"));
+                return null;
+            }
+            else if (locationLine.Start.Z > locationLine.End.Z)
+            {
+                BH.Engine.Base.Compute.RecordNote(string.Format("The input line's bottom was above its top. This line has been flipped to resolve the issue."));
+                locationLine = locationLine.Flip();
+            }
+            return locationLine;
+        }
```
vietle-bh, line 52: Test

Revit_Engine/Modify/SortEndpoints.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            if (locationLine.Start.Z == locationLine.End.Z)
+            {
+                BH.Engine.Base.Compute.RecordError(string.Format("Line's start and end points have the same elevation.}"));
+                return null;
+            }
+            else if (locationLine.Start.Z > locationLine.End.Z)
+            {
+                BH.Engine.Base.Compute.RecordNote(string.Format("The input line's bottom was above its top. This line has been flipped to resolve the issue."));
+                locationLine = locationLine.Flip();
```
vietle-bh, line 49:
```suggestion
int found = 1;
```

Revit_Engine/Modify/SortEndpoints.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            return locationLine;
```
vietle-bh, line 51: Test

Revit_Engine/Modify/SortEndpoints.cs
vietle-bh, line 52: Thanks @pawelbaran, I've made the changes but since we shouldn't modify the original Bhom column object's location line, I needed to add the same geometry check to `SetLocation`. This led to duplicated messages:

image

Can we instead separate the checking of line endpoints to a new `SortEndpoints` method under Revit_Engine\Modify? I can then call it from both `SetLocation` and `FamilyInstance`?

pawelbaran: I see two options here @vietle-bh:
- Use exactly same note in both create and update (identical warnings are suppressed, only one is shown in the UI)
- Create a new method as you suggest - but then please be mindful that it needs to highlight the fact that it is used to sort column's end points to avoid misuse (and the best way to highlight is to embed the information in the method name). Also I would rather put the method in Revit_Core_Engine instead of Revit_Engine - the latter is reflected in all UIs, resulting in the internally used method being available for example in GH.

Both solutions work for me, please feel free to choose the one you prefer 👍

vietle-bh: Thanks @pawelbaran! I implemented the 2nd option in an attempt to stay DRY :relieved:

Revit_Core_Engine/Modify/SortColumEndpoints.cs
```diff
@@ -25,34 +25,33 @@
...
         /***************************************************/
 
-        [Description("Flip the input line if its start point is higher than its endpoint.")]
-        [Input("line", "Line to check.")]
-        [Output("line", "The original line if its start point is already below its endpoint. Otherwise, the line flipped.")]
-        public static Line SortEndpoint(this Line locationLine)
+        [Description("Flip the input line of a BHoM Column object if its start point is higher than its endpoint. This allows pushing it to Revit")]
+        [Input("line", "Column line to check.")]
+        [Output("line", "The original column line if its start point is already below its endpoint. Otherwise, the same line flipped.")]
+        public static Line SortColumEndpoints(this Line columnLine)
```
pawelbaran, line 39: Both file and method names have a typo 🙈 

Besides that welcome to the swamp of a large code base! With your current implementation, there would be 2 errors in case of both points on the same elevation:
- one on the level of this method, with no info about the element it refers to
- one on the convert level, with the info about the element it refers to

Not ideal. Since you took this slightly more tricky route, I would go a step further now and rename the method to something like `BH.Revit.Engine.Core.Query.ColumnCurve`, taking the `IFramingElement` and returning the line. Inside you would do all necessary checks: if the framing element has location, if the location is a line, if it is flipped etc., everything that is currently done in the convert method. Then you could also manage the errors/warnings/notes better, plus unify the inputs for both . What do you think?

Revit_Core_Engine/Modify/SortColumEndpoints.cs
```diff
@@ -25,34 +25,33 @@
...
         {
-            if (locationLine.Start.Z == locationLine.End.Z)
+            if (columnLine.Start.Z == columnLine.End.Z)
```
pawelbaran, line 41: We should always include tolerance in geometrical queries to avoid numerical noise and negligible inaccuracies. So if you want to compare 2 geometrical objects for equality of any type, it is worth using `BH.oM.Geometry.Tolerance.Distance` (or others from this class):
```suggestion
            if (Math.Abs(columnLine.Start.Z - columnLine.End.Z) <= BH.oM.Geometry.Tolerance.Distance)
```

pawelbaran (changes requested):
This escalated quickly 😃 Please lemme know if you need a chat to discuss my comments.

FraserGreenroyd (changes requested):

Revit_Core_Engine/Query/ColumnLine.cs
```diff
@@ -0,0 +1,68 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Extracts the location line of a BHoM Column object.")]
+        [Input("column", "A BHoM Column object.")]
+        [Output("line", "The location line of the BHoM Column object.")]
+        public static Line ColumnLine(this Column column)
+        {
+            Line columnLine = (BH.oM.Geometry.Line)column.Location;
```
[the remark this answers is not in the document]
vietle-bh: Thanks. I've added the null handler now.

Revit_Core_Engine/Query/ColumnLine.cs
```diff
@@ -0,0 +1,68 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Extracts the location line of a BHoM Column object.")]
+        [Input("column", "A BHoM Column object.")]
+        [Output("line", "The location line of the BHoM Column object.")]
+        public static Line ColumnLine(this Column column)
```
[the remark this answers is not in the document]
vietle-bh: I don't see any existing method in the Revit_Toolkit or `BHoM_Engine.Physical_Engine`...
@pawelbaran, do you see one?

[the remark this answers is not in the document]
pawelbaran: Not really, I think this one is very Revit-specific, Revit only allows lines with startpoint below the endpoint, while in the wider BHoM we accept any sort of curves for columns: nonlinear ones, flipped etc.

Revit_Core_Engine/Query/ColumnLine.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            if (column == null)
+            {
+                BH.Engine.Base.Compute.RecordError(string.Format("Cannot read location line because column is null. BHoM_Guid: {0}", column.BHoM_Guid));
+                return null;
+            }
+
+            Line columnLine = (BH.oM.Geometry.Line)column.Location;
```
pawelbaran, line 49: This will crash if the location is not a `Line`, but, say, an `Arc`. This should be written as follows:
```suggestion
            Line columnLine = column.Location as BH.oM.Geometry.Line;
```

Revit_Core_Engine/Query/ColumnLine.cs
```diff
@@ -0,0 +1,73 @@
...
+        /***************************************************/
+
+        [Description("Extracts the location line of a BHoM Column object.")]
+        [Input("column", "A BHoM Column object.")]
+        [Output("line", "The location line of the BHoM Column object.")]
+        public static Line ColumnLine(this Column column)
+        {
+            if (column == null)
+            {
+                BH.Engine.Base.Compute.RecordError(string.Format("Cannot read location line because column is null. BHoM_Guid: {0}", column.BHoM_Guid));
```
pawelbaran, line 45: These days I personally I prefer string formatting based on `$""` syntax instead of the good old `string.Format`. Please feel free to stick to the current version if this is your preference, but if not then I would recommend using the more contemporary syntax - it makes the code a bit more readable.

Just in case, this is a general comment to all code that you will be writing, but again, please be aware that this is just my preference and we are all allowed to have our own 😉

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -84,17 +84,15 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
                 return null;
             }
 
-            if (((BH.oM.Geometry.Line)framingElement.Location).Start.Z >= ((BH.oM.Geometry.Line)framingElement.Location).End.Z)
-            {
-                BH.Engine.Base.Compute.RecordError(string.Format("Start point of Revit columns need to have lower elevation than the end point. Have a look at flipping your location curves. BHoM_Guid: {0}", framingElement.BHoM_Guid));
+            BH.oM.Geometry.Line locationLine = framingElement.ColumnLine();
```
pawelbaran, line 87: There is no need for the two checks above (`if (framingElement.Location == null)` and `if (framingElement.Location as BH.oM.Geometry.Line == null)` as you are checking this inside `ColumnLine` method.

pawelbaran (changes requested):
I think we are getting there, but a few minor tweaks are still needed.

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
vietle-bh: Thanks for the great tip! I mistakenly thought these methods did the same thing :scream:

Revit_Core_Engine/Query/ColumnLine.cs
FraserGreenroyd: I would second the use of  `$""`! :smile:

pawelbaran: Personally I was considering a dedicated push to replace all `String.Format` with `$""` across the entire BIM-related code base...

FraserGreenroyd: That wouldn't be such a bad idea for consistency @pawelbaran - could be a nice easy compliance task for someone to learn more about the code base... 😉

pawelbaran: Man, I would write a script for it 😃

FraserGreenroyd: Just be careful not to fall foul of the automated code modification regulations set out in our BHoMBot procedures then 😉

pawelbaran: Did someone say foul?

image

Revit_Core_Engine/Query/ColumnLine.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Extracts the location line of a BHoM Column object.")]
+        [Input("column", "A BHoM Column object.")]
+        [Output("line", "The location line of the BHoM Column object.")]
```
pawelbaran, line 40: Not sure if that has been stressed enough in the onboarding sessions @vietle-bh, but we are trying to be specific in the descriptions, assuming that it is worth spending a bit more time on creation to avoid confusion in the future. As long as it is not always the case for the existing code base (unfortunately), I think it is a good custom when creating new stuff. I have made some suggestions below, but feel free to phrase it as you like - I am happy as long as the information is conveyed in an understandable form :wink:

```suggestion
        [Description("Extracts the location line of a BHoM Column object in preparation to push to Revit. Includes validity checks and flipping reversed nodes.")]
        [Input("column", "A BHoM Column object to extract the line from.")]
        [Output("line", "Preprocessed location line of the BHoM Column object to be used on push to Revit.")]
```

pawelbaran (changes requested):
Last bit 😃 Complaining as much as I can on this one @vietle-bh, so that you know what to look at in the future. But in general looks good, I tested and am happy with the code, well done!

Revit_Core_Engine/Query/ColumnLine.cs
vietle-bh: Thanks. I'll make sure to do the same for future descriptions!

Revit_Core_Engine/Query/ColumnLine.cs
FraserGreenroyd: Nothing is fouler than the code contained in any BIM related repository...

image

pawelbaran (approved):
Happy with the changes now, reviewed and tested. Well done!

pawelbaran:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check serialisation
@BHoMBot check installer
@BHoMBot check versioning
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 6006585507

PR #1200, opened 2022-04-28
RevitComparisonConfig.NumericTolerance overridden

### LINKED ISSUE #1199: Override RevitComparisonConfig.NumericTolerance to 1e-6
#### Description:

Default numeric tolerance value is `Double.MinValue`. As discussed with @alelom, for practical Revit purposes it is way too precise, with 1e-6 being a better choice as a default option.

pawelbaran:

Closes #1199

### Test files

Invoke the RevitDiffing tests in [DiffingTests_Prototypes ](https://github.com/BHoM/DiffingTests_Prototypes). They should all pass.

### Changelog

- The default NumericalTolerance for RevitDiffing is now set to **`1e-6`** instead of `Double.MaxValue`. This will affect every Diffing done with RevitDiffing from now on. Comparisons on the same sets of objects done previously with RevitDiffing will return different results after this PR, if the default tolerance of was used. To replicate the same results as before, set `Double.MaxValue` manually on the NumericTolerance property of the RevitComparisonConfig before triggering the RevitDiffing.

### DIFF
--- a/Revit_oM/Config/RevitComparisonConfig.cs
+++ b/Revit_oM/Config/RevitComparisonConfig.cs
@@ -68,6 +68,12 @@ namespace BH.oM.Adapters.Revit
         [Description("(Defaults to `true`) If false, if an object has a RevitParameter with a null Value deleted from it, then the owner object is NOT considered 'Modified' and the Comparison will NOT return this difference.")]
         public virtual bool RevitParams_ConsiderRemovedUnassigned { get; set; } = true;
 
+        [Description("Numeric tolerance for property values, applied to all numerical properties. Defaults to 1e-6." +
+                     "\nApplies rounding for numbers smaller than this." +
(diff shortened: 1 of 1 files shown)


alelom (approved):
This PR was tested with the DiffingTests_Prototypes toolkit. No errors were reported by it.

All scripts are bound to keep working as before, with the exception that the results returned by the RevitDiffing (on the same couples of input object sets) may be different because of this PR.

This approach is the preferred one when modifying default values for ComparisonConfigs (and finally a good use case for `virtual`/`override`)! 😄

pawelbaran:

pawelbaran:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 6225050990

PR #1203, opened 2022-05-25
Get the control points of the Curve Loop

### LINKED ISSUE #1202: Get the control points of the Curve Loop
#### Description:

As in the title.

michal-pekacki:

Closes #1202

Method to get the control points of the `CurveLoop` object.

### Additional comments

This PR has been closed ([New PR for the method](https://github.com/BHoM/Revit_Toolkit/pull/1230))

### DIFF
--- a/Revit_Core_Engine/Query/Points.cs
+++ b/Revit_Core_Engine/Query/Points.cs
@@ -1,0 +1,74 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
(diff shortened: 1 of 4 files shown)


michal-pekacki:
@BHoMBot check required

vietle-bh (approved):
All good. No comment from me :blush:

michal-pekacki:
@BHoMBot check ready-to-merge

Revit_Core_Engine/Query/Points.cs
```diff
@@ -0,0 +1,74 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns control points of the CurveLoop.")]
+        [Input("curveLoop", "CurveLoop to get the points from.")]
+        [Output("points", "List of uniqe curve loop points")]
+        public static List<XYZ> Points(this CurveLoop curveLoop)
```
pawelbaran, line 39: I would call this method `Tessellate` and write it along these lines:
```suggestion
        public static List<XYZ> Tessellate(this CurveLoop curveLoop)
        {
            List<XYZ> result = new List<XYZ>();
            foreach (Curve curve in curveLoop)
            {
                IList<XYZ> points = curve.Tessellate();
                if (points.Count != 0)
                    result.AddRange(points.Take(points.Count - 1));
            }
            
            if (curveLoop.IsOpen())
                result.Add(curveLoop.Last().GetEndPoint(1));
            
            return result;
        }
```

Wouldn't that return the tessellated loop? I think this would be more correct than your solution because its order will not be distorted by culling duplicated (which can always be done separately if needed).

*I typed the above on the web page so the code may not compile, just showing the concept

pawelbaran (changes requested):

Revit_Core_Engine/Query/Points.cs
michal-pekacki: Good idea - to be done in the new branch.

PR #1218, opened 2022-07-01
Scenario filtering empty parameter fix

### LINKED ISSUE #1217: Scenario filtering empty parameter bug
#### Description:

Filtering doesn't work correctly for empty parameter values.

#### Steps to reproduce:

1. Open Insulation Tool setting window
2. Set filtering in scenario for any parameter as Not Equal or Not Contains (may be Comments)
3. Run apply method
4. All elements with the empty parameter value will be out of the scenario what is wrong

#### Expected behaviour:

Elements with empty values to pass scenario filtering.

michal-pekacki:

Closes #1217 

Fixed `Passes` method to work with empty parameter values correctly.

### Test files

[Revit 2018 model](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BuroHappoldEngineering/BuroHappold_Revit_Toolkit/%23461-InsulationToolQA?csf=1&web=1&e=l586BA)

1. Open Insulation Tool setting window
2. Set filtering in scenario for any parameter as Not Equal or Not Contains (may be Comments)
3. Run apply method

### DIFF
--- a/Revit_Core_Engine/Query/Passes.cs
+++ b/Revit_Core_Engine/Query/Passes.cs
@@ -199,7 +199,7 @@ namespace BH.Revit.Engine.Core
             if (!CheckIfNotNull(parameter, request))
                 return false;
 
-            if (parameter.HasValue && parameter.StorageType == StorageType.Integer && parameter.Definition.ParameterType == ParameterType.YesNo)
+            if (parameter.StorageType == StorageType.Integer && parameter.Definition.ParameterType == ParameterType.YesNo)
             {
                 int paramValue = parameter.AsInteger();
                 return (request.Value && paramValue == 1) || (!request.Value && paramValue == 0);
@@ -253,9 +253,10 @@ namespace BH.Revit.Engine.Core
             if (!CheckIfNotNull(parameter, request))
                 return false;
 
+
+            double paramValue;
             if (parameter.HasValue)
             {
-                double paramValue;
                 if (parameter.StorageType == StorageType.Double)
                     paramValue = parameter.AsDouble();
                 else if (parameter.StorageType == StorageType.Integer)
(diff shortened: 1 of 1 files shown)


Revit_Core_Engine/Query/Passes.cs
```diff
@@ -199,7 +199,7 @@ public static bool Passes(this Parameter parameter, FilterByParameterBool reques
             if (!CheckIfNotNull(parameter, request))
                 return false;
 
-            if (parameter.HasValue && parameter.StorageType == StorageType.Integer && parameter.Definition.ParameterType == ParameterType.YesNo)
+            if (parameter.StorageType == StorageType.Integer && parameter.Definition.ParameterType == ParameterType.YesNo)
```
pawelbaran, line 202: I would not agree with this change - no value is not equal to `false`. After your change, the empty parameter will pass a request asking for values equal to `false`, which is not true. However, correct me if I am wrong: this will never occur in practice anyways, taken that yes/no can't be empty in Revit, right?

Revit_Core_Engine/Query/Passes.cs
```diff
@@ -253,9 +253,10 @@ public static bool Passes(this Parameter parameter, FilterByParameterNumber requ
             if (!CheckIfNotNull(parameter, request))
                 return false;
 
+
+            double paramValue;
```
pawelbaran, line 257: You can get rid of the `else` statement easily
```suggestion
            double paramValue = double.NaN;
```

Revit_Core_Engine/Query/Passes.cs
```diff
@@ -267,31 +268,38 @@ public static bool Passes(this Parameter parameter, FilterByParameterNumber requ
...
-                        return paramValue - comparisonValue < -comparisonTolerance;
-                    case NumberComparisonType.LessOrEqual:
-                        return paramValue - comparisonValue < comparisonTolerance;
-                    case NumberComparisonType.NotEqual:
-                        return Math.Abs(paramValue - comparisonValue) > comparisonTolerance;
-                }
+            switch (request.NumberComparisonType)
+            {
+                case NumberComparisonType.Equal:
+                    return Math.Abs(paramValue - comparisonValue) <= comparisonTolerance;
```
pawelbaran, line 289: One more edge case to cover:
```suggestion
                    return Math.Abs(paramValue - comparisonValue) <= comparisonTolerance || (double.IsNaN(paramValue) && double.IsNaN(comparisonValue));
```

pawelbaran (changes requested):
Good tweaks in general, I may have found a few further improvements - see the code comments

Revit_Core_Engine/Query/Passes.cs
pawelbaran: Hmm, if it was an academic discussion, I would argue that not checked != unchecked 🐒 But I understand your approach from the user's perspective, so happy to agree with you 👍

pawelbaran (approved):
Happy to approve the changes now

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check versioning
@BHoMBot check ready-to-merge

PR #1224, opened 2022-07-13
DoesIntersect method for checking the intersection between elements

### LINKED ISSUE #1123: Add method to query the closest element from another element
#### Description:
Suggesting new method to query what is the closest element of a certain category to another element, enabling searching on linked files.

### LINKED ISSUE #1223: Methods for checking the intersection between elements
#### Description:

Methods for checking the intersection between:
- Host vs Host elements
- Host vs Link elements
- Link vs Link elements
- BBox vs Host elements
- BBox vs Link elements

michal-pekacki:

Closes #1223 

- Added `Does Intersect` method for checking the intersection between elements.
- Method `Bounds` for getting the bounding box of the element has been changed.
- Extend `ToSolid` method for bounding box elements.

### Test files

[Revit 2018 models](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/%231224-DoesIntersect?csf=1&web=1&e=eJhNxD)
- `MEP Model` as a Host
- `Arch Model` as a Link
- `LinkVsLink` to check intersection between linked elements

### DIFF
--- a/Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
+++ b/Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
@@ -97,7 +97,7 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         [Description("Converts Revit bounding box to a Revit Solid.")]
-        [Input("scopeBox", "Revit bounding box to be converted.")]
+        [Input("bbox", "Revit bounding box to be converted.")]
         [Input("settings", "Revit adapter settings to be used while performing the convert.")]
         [Output("solid", "Revit Solid resulting from converting the input Revit scope box.")]
         public static Solid ToSolid(this BoundingBoxXYZ bbox, RevitSettings settings = null)
--- a/Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
+++ b/Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
@@ -95,6 +95,40 @@ namespace BH.Revit.Engine.Core
         }
 
         /***************************************************/
+
+        [Description("Converts Revit bounding box to a Revit Solid.")]
+        [Input("scopeBox", "Revit bounding box to be converted.")]
+        [Input("settings", "Revit adapter settings to be used while performing the convert.")]
+        [Output("solid", "Revit Solid resulting from converting the input Revit scope box.")]
+        public static Solid ToSolid(this BoundingBoxXYZ bbox, RevitSettings settings = null)
+        {
+            XYZ pt0 = new XYZ(bbox.Min.X, bbox.Min.Y, bbox.Min.Z);
+            XYZ pt1 = new XYZ(bbox.Max.X, bbox.Min.Y, bbox.Min.Z);
+            XYZ pt2 = new XYZ(bbox.Max.X, bbox.Max.Y, bbox.Min.Z);
+            XYZ pt3 = new XYZ(bbox.Min.X, bbox.Max.Y, bbox.Min.Z);
+
+            Line edge0 = Line.CreateBound(pt0, pt1);
+            Line edge1 = Line.CreateBound(pt1, pt2);
+            Line edge2 = Line.CreateBound(pt2, pt3);
+            Line edge3 = Line.CreateBound(pt3, pt0);
+
+            List<Curve> edges = new List<Curve>();
+            edges.Add(edge0);
+            edges.Add(edge1);
+            edges.Add(edge2);
+            edges.Add(edge3);
+
+            double height = bbox.Max.Z - bbox.Min.Z;
+
+            CurveLoop baseLoop = CurveLoop.Create(edges);
+
+            List<CurveLoop> loopList = new List<CurveLoop>();
+            loopList.Add(baseLoop);
+
+            Solid solid = GeometryCreationUtilities.CreateExtrusionGeometry(loopList, XYZ.BasisZ, height);
+            return SolidUtils.CreateTransformed(solid, bbox.Transform);
+        }
+
     }
 }
 
--- a/Revit_Core_Engine/Query/Bounds.cs
+++ b/Revit_Core_Engine/Query/Bounds.cs
@@ -55,13 +55,11 @@ namespace BH.Revit.Engine.Core
 
             foreach (Solid solid in solids)
             {
-                BoundingBoxXYZ solidBBox = new BoundingBoxXYZ();
-
                 Solid newSolid = solid;
                 if (transform != null)
                     newSolid = SolidUtils.CreateTransformed(solid, transform);
 
-                solidBBox = newSolid.GetBoundingBox();
+                BoundingBoxXYZ solidBBox = newSolid.GetBoundingBox();
 
                 XYZ solidMin = solidBBox.Min;
                 XYZ solidMax = solidBBox.Max;
--- a/Revit_Core_Engine/Query/Bounds.cs
+++ b/Revit_Core_Engine/Query/Bounds.cs
@@ -57,15 +57,11 @@ namespace BH.Revit.Engine.Core
             {
                 BoundingBoxXYZ solidBBox = new BoundingBoxXYZ();
 
+                Solid newSolid = solid;
                 if (transform != null)
-                {
-                    Solid transSolid = SolidUtils.CreateTransformed(solid, transform);
-                    solidBBox = transSolid.GetBoundingBox();
-                }
-                else
-                {
-                    solidBBox = solid.GetBoundingBox();
-                }
+                    newSolid = SolidUtils.CreateTransformed(solid, transform);
+
+                solidBBox = newSolid.GetBoundingBox();
 
                 XYZ solidMin = solidBBox.Min;
                 XYZ solidMax = solidBBox.Max;
--- a/Revit_Core_Engine/Query/Bounds.cs
+++ b/Revit_Core_Engine/Query/Bounds.cs
@@ -55,7 +55,18 @@ namespace BH.Revit.Engine.Core
 
             foreach (Solid solid in solids)
             {
-                BoundingBoxXYZ solidBBox = solid.GetBoundingBox();
+                BoundingBoxXYZ solidBBox = new BoundingBoxXYZ();
+
+                if (transform != null)
+                {
+                    Solid transSolid = SolidUtils.CreateTransformed(solid, transform);
+                    solidBBox = transSolid.GetBoundingBox();
+                }
+                else
+                {
+                    solidBBox = solid.GetBoundingBox();
+                }
+
                 XYZ solidMin = solidBBox.Min;
                 XYZ solidMax = solidBBox.Max;
                 XYZ solidOrigin = solidBBox.Transform.Origin;
@@ -71,12 +82,6 @@ namespace BH.Revit.Engine.Core
             BoundingBoxXYZ unionBbox = new BoundingBoxXYZ();
             unionBbox.Min = new XYZ(minX, minY, minZ);
             unionBbox.Max = new XYZ(maxX, maxY, maxZ);
-
-            if (transform != null)
-            {
-                Solid union = SolidUtils.CreateTransformed(unionBbox.ToSolid(), transform);
-                unionBbox = union.GetBoundingBox();
-            }
 
             return unionBbox;
         }
--- a/Revit_Core_Engine/Query/Bounds.cs
+++ b/Revit_Core_Engine/Query/Bounds.cs
@@ -73,7 +73,10 @@ namespace BH.Revit.Engine.Core
             unionBbox.Max = new XYZ(maxX, maxY, maxZ);
 
             if (transform != null)
-                unionBbox.Transform = transform;
+            {
+                Solid union = SolidUtils.CreateTransformed(unionBbox.ToSolid(), transform);
+                unionBbox = union.GetBoundingBox();
+            }
 
             return unionBbox;
         }
--- a/Revit_Core_Engine/Query/Bounds.cs
+++ b/Revit_Core_Engine/Query/Bounds.cs
@@ -22,6 +22,7 @@
 
 using Autodesk.Revit.DB;
 using BH.oM.Base.Attributes;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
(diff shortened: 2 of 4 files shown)


Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,176 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Check if two elements intersect.")]
+        [Input("element1", "First element to check the intersection for.")]
+        [Input("element2", "Second element to check the intersection for.")]
+        [Output("bool", "Result of the intersect checking.")]
+        public static bool DoesIntersect(this Element element1, Element element2)
+        {
```
pawelbaran, line 43: It tempts me to day that this method should be as simple as this (for all cases link vs host etc.), more or less (code typed by hand):
```C#
        Transform transform1 = element1.Document.LinkTransform(); //Null if doc not linked
        Transform transform2 = element2.Document.LinkTransform();
        List<Solid> solids = element1.Solids();
        if (transform1 != null)
            solids = solids.Transform(transform1); //There is no such a method, but ykwim

        if (transform2 != null)
            solids = solids.Transform(transform2.Inverse);

        //Possibly some more pre-checks with bounding boxes here to improve performance
        foreach (Solid solid in solids)
        {
            ElementIntersectsSolidFilter intersectFilter = new ElementIntersectsSolidFilter(solid);
            if (intersectFilter.PassesFilter(element2))
                return true;
        }

I may be wrong but it sounds as if it could be generalised a lot
```

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,176 @@
...
+                List<Solid> element1Solids = element1.Solids(new Options());
+                List<Solid> element2Solids = element2.Solids(new Options());
+                List<Solid> el1TransSolids = element1Solids.Select(x => SolidUtils.CreateTransformed(x, transform1)).ToList();
+                List<Solid> el2TransSolids = element2Solids.Select(x => SolidUtils.CreateTransformed(x, transform2)).ToList();
+
+                foreach (Solid el1Solid in el1TransSolids)
+                {
+                    foreach (Solid el2Solid in el2TransSolids)
+                    {
+                        Solid intersection = BooleanOperationsUtils.ExecuteBooleanOperation(el1Solid, el2Solid, BooleanOperationsType.Intersect);
```
pawelbaran, line 122: Isn't `ElementIntersectsSolid` precise enough?

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,176 @@
...
+            return false;
+        }
+
+        /***************************************************/
+
+        [Description("Check if bounding box intersects with element.")]
+        [Input("bbox", "Bounding box to check the intersection for.")]
+        [Input("element", "Element to check the intersection for.")]
+        [Output("bool", "Result of the intersect checking.")]
+        public static bool DoesIntersect(this BoundingBoxXYZ bbox, Element element)
```
pawelbaran, line 138: This method reminds me of my first comment about the risk of bounding box transform being forgotten 😃

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                return false;
+            }
+
+            Document doc = element.Document;
+
+            Outline outline = new Outline(bbox.Min, bbox.Max);
+            BoundingBoxIntersectsFilter bboxIntersect = new BoundingBoxIntersectsFilter(outline);
+            BoundingBoxIsInsideFilter bboxInside = new BoundingBoxIsInsideFilter(outline);
+            LogicalOrFilter bboxFilter = new LogicalOrFilter(new List<ElementFilter> { bboxIntersect, bboxInside });
```
pawelbaran, line 150: This part could be inside the first `if`

pawelbaran (changes requested):

Revit_Core_Engine/Query/Bounds.cs
```diff
@@ -45,16 +46,36 @@ public static BoundingBoxXYZ Bounds(this IEnumerable<Solid> solids, Transform tr
...
+            BoundingBoxXYZ unionBbox = new BoundingBoxXYZ();
+            unionBbox.Min = new XYZ(minX, minY, minZ);
+            unionBbox.Max = new XYZ(maxX, maxY, maxZ);
+
             if (transform != null)
-                union = SolidUtils.CreateTransformed(union, transform);
+                unionBbox.Transform = transform;
 
-            return union.GetBoundingBox();
+            return unionBbox;
```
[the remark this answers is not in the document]
michal-pekacki: Yes that's right. I will change that to keep the previous way (transform the solid -> convert to bounding box) to avoid different coordinate systems and issues with bounding box transform.

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            Document doc1 = element1.Document;
+            Document doc2 = element2.Document;
+
+            //host vs host
+            if (!doc1.IsLinked && !doc2.IsLinked)
+            {
+                BoundingBoxXYZ bbox1 = element1.PhysicalBounds();
+                BoundingBoxXYZ bbox2 = element2.PhysicalBounds();
```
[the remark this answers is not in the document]
michal-pekacki: Good idea for method optimization 👍

Revit_Core_Engine/Query/Bounds.cs
```diff
@@ -45,16 +46,39 @@ public static BoundingBoxXYZ Bounds(this IEnumerable<Solid> solids, Transform tr
...
+            BoundingBoxXYZ unionBbox = new BoundingBoxXYZ();
+            unionBbox.Min = new XYZ(minX, minY, minZ);
+            unionBbox.Max = new XYZ(maxX, maxY, maxZ);
+
             if (transform != null)
-                union = SolidUtils.CreateTransformed(union, transform);
+            {
+                Solid union = SolidUtils.CreateTransformed(unionBbox.ToSolid(), transform);
+                unionBbox = union.GetBoundingBox();
+            }
```
pawelbaran, line 79: I am afraid this may not be correct: in your case, the resultant bounding box will be red (bounding box of a green bounding box), while it should be blue.

image

Instead, I would rather transform each solid inside the `foreach` loop.

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (element1 == null || element2 == null)
+            {
+                return false;
+            }
+
+            Transform transform1 = element1.Document.LinkTransform();
+            Transform transform2 = element2.Document.LinkTransform();
+
+            //host vs host
+            if (transform1 == null && transform2 == null)
```
pawelbaran, line 53: Worth checking if transform is `null` or identity - this means the same in this particular context

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                ElementIntersectsElementFilter intersectFilter = new ElementIntersectsElementFilter(element2);
+                return intersectFilter.PassesFilter(element1);
+            }
+            //host vs link || link vs host
+            else if ((transform1 == null && transform2 != null) || (transform1 != null && transform2 == null))
```
pawelbaran, line 59: The code is still far too complicated. What you can do is to simplify the problem by finding:
- element staying untouched, to be checked against `ElementIntersectsSolidFilter` (host element, if both elements linked then the 1st one)
- element to be turned into a solid and transformed (linked element, if both elements linked then the 2nd one)
- transform to apply to the extracted solids (either 1 transform or 1 multiplied by inverse of 2, if both elements are linked)

So in the end, you will have 2 scenarios:
a) both link transforms null or identity
b) at least one not meeting requirements of a), then identify the 3 objects mentioned above and run the query

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,170 @@
...
+        [Input("element", "Element to check the intersection for.")]
+        [Output("bool", "Result of the intersect checking.")]
+        public static bool DoesIntersect(this BoundingBoxXYZ bbox, Element element)
+        {
+            if (bbox == null || element == null)
+            {
+                return false;
+            }
+
+            if (!bbox.Transform.IsIdentity)
```
pawelbaran, line 100: I guess this could be relaxed by allowing translation, so `bbox.Transform.IsTranslation` and then adding the origin to the min/max of the outline. Or would that be problematic by any means?

pawelbaran (changes requested):

Revit_Core_Engine/Query/Bounds.cs
michal-pekacki: Yes, that still need to be corrected. Thanks! 👍

Revit_Core_Engine/Query/Bounds.cs
```diff
@@ -45,16 +46,44 @@ public static BoundingBoxXYZ Bounds(this IEnumerable<Solid> solids, Transform tr
...
+
+                if (transform != null)
+                {
+                    Solid transSolid = SolidUtils.CreateTransformed(solid, transform);
+                    solidBBox = transSolid.GetBoundingBox();
+                }
+                else
+                {
+                    solidBBox = solid.GetBoundingBox();
+                }
```
pawelbaran, line 68: This could be done in 4 lines:
```suggestion
                if (transform != null)
                    solid = SolidUtils.CreateTransformed(solid, transform);
                
                BoundingBoxXYZ solidBBox = solid.GetBoundingBox();
```

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,156 @@
...
+                else if (!transform1.IsIdentity && transform2.IsIdentity)
+                {
+                    return DoesIntersectWithTransform(element2, element1, transform1);
+                }
+                //link vs link
+                if (!transform1.IsIdentity && !transform2.IsIdentity)
+                {
+                    Transform doubleTransform = transform2.Multiply(transform1.Inverse);
+                    return DoesIntersectWithTransform(element1, element2, doubleTransform);
+                }
```
pawelbaran, line 75: My gut feeling tells me this could be written in 2 lines:
```suggestion
                Transform doubleTransform = transform2.Multiply(transform1.Inverse);
                return DoesIntersectWithTransform(element1, element2, doubleTransform);
```

Revit_Core_Engine/Query/DoesIntersect.cs
pawelbaran: I think you need to change this line too (please make sure that identity transform returns `IsTranslation == true`).
```suggestion
            if (!bbox.Transform.IsTranslation)
```

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,156 @@
...
+        public static bool DoesIntersect(this BoundingBoxXYZ bbox, Element element)
+        {
+            if (bbox == null || element == null)
+            {
+                return false;
+            }
+
+            if (!bbox.Transform.IsIdentity)
+            {
+                BH.Engine.Base.Compute.RecordWarning("Intersection of the bounding boxe could not be checked. Only identity transformation is currently supported.");
```
pawelbaran, line 96:
```suggestion
                BH.Engine.Base.Compute.RecordWarning("Intersection of the bounding boxes could not be checked. Only identity transformation is currently supported.");
```

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,156 @@
...
+        /***************************************************/
+
+        [Description("Check if two elements intersect.")]
+        [Input("hostElement", "First element to check the intersection for.")]
+        [Input("transElement", "Second element to check the intersection for.")]
+        [Input("transform", "Transformation of the second element.")]
+        [Output("bool", "Result of the intersect checking.")]
+        private static bool DoesIntersectWithTransform(this Element hostElement, Element transElement, Transform transform)
+        {
+
```
pawelbaran, line 140: redundant blank

Revit_Core_Engine/Query/DoesIntersect.cs
```diff
@@ -0,0 +1,156 @@
...
+            List<Solid> solids = transElement.Solids(new Options()).Select(x => SolidUtils.CreateTransformed(x, transform)).ToList();
```
pawelbaran, line 141: `CreateTransformed` could be moved inside the `foreach` loop - not sure if very relevant, but can imagine it improving the performance by a few % by not doing the transformation on all geometries, only the ones before the intersection

pawelbaran (changes requested):
@michal-pekacki you will hate me for this PR, but once you start hating, please see the first and last version of the code - I think the improvement is worth it 💪

Revit_Core_Engine/Query/DoesIntersect.cs
michal-pekacki: Yes, my mistake here. Thanks! 👍

Revit_Core_Engine/Query/DoesIntersect.cs
michal-pekacki: Wow, I see that 😃

Revit_Core_Engine/Query/Bounds.cs
```diff
@@ -45,16 +46,40 @@ public static BoundingBoxXYZ Bounds(this IEnumerable<Solid> solids, Transform tr
...
+            double minZ = double.MaxValue;
+            double maxX = double.MinValue;
+            double maxY = double.MinValue;
+            double maxZ = double.MinValue;
+
+            foreach (Solid solid in solids)
             {
-                union = BooleanOperationsUtils.ExecuteBooleanOperation(union, solid, BooleanOperationsType.Union);
+                BoundingBoxXYZ solidBBox = new BoundingBoxXYZ();
+
```
pawelbaran, line 59: These two lines could be gone

Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
```diff
@@ -95,6 +95,40 @@ public static Solid ToSolid(this Element scopeBox, RevitSettings settings = null
...
+
+            double height = bbox.Max.Z - bbox.Min.Z;
+
+            CurveLoop baseLoop = CurveLoop.Create(edges);
+
+            List<CurveLoop> loopList = new List<CurveLoop>();
+            loopList.Add(baseLoop);
+
+            Solid solid = GeometryCreationUtilities.CreateExtrusionGeometry(loopList, XYZ.BasisZ, height);
+            return SolidUtils.CreateTransformed(solid, bbox.Transform);
```
pawelbaran, line 129: Sanity check - I just looked up the API description and it sounds a bit like you would need to take inverse transform here... Have you tested this bit?
image

pawelbaran (commented):
The code looks good now - last two comments 👍

Revit_Core_Engine/Convert/Revit/Internal/ToSolid.cs
pawelbaran: After a discussion with @michal-pekacki we came to a conclusion that the code is correct, while the API docs are a bit ambiguous.

michal-pekacki: I tested that and it works correctly.

pawelbaran (approved):
Happy with the code now, really well done @michal-pekacki - looking forward to doing more optimisation around this sort of filtering in the future! 💪

pawelbaran:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check required

PR #1228, opened 2022-07-18
Revit_Engine_Core-#1227-Query methods for element ownership status

### LINKED ISSUE #1227: Element Checkout Status Query
#### Description:

Ability to check and elements worksharing ownership/checkout status to determine which elements are immediately editable, and which require other users to relinquish ownsership.

travispotterBH:

Closes #1227

Methods have been added to support the checking of single and lists of elements for their ownership/checkout status. 

### Changelog

- `Query.ElementsOwnedByOtherUsers()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.ElementsOwnedByCurrentUser()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.ElementsOwnedByNone()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.ElementsEditablePerCheckoutStatus()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.IsOwnedByOtherUser()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.IsOwnedByCurrentUser()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.IsOwnedByNone()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Query.CheckoutStatus()` Query method added in the `Revit_Core_Engine` for `Element` class
- `Modify.Checkout()` Modify method added in the `Revit_Core_Engine` for `Element` class
- `ElementOwnedByOtherUserWarning()` Warning added to `Revit_Core_Engine` for `Element` class
- `ElementOwnedByCurrentUserNote()` Note added to `Revit_Core_Engine` for `Element` class

### Additional comments

Note: The methods work over a cached local version to determine ownership. They will only be as up to date at the most recently sync + reload by the user. 
https://www.revitapidocs.com/2023/1a4b4bbb-060d-1f42-fbb2-ab85081f8e7f.htm
image
I attempted to prompted the user to reload latest via task dialog window which alerted the user to the need to reload and asked for confirmation to proceed with a reload latest. However, the reload latest fails when called from one of the methods due to an open transaction. We could possibly set some rules on the toolkits Run command for methods that modify elements to do a reload latest before initiating the transaction as a resolution to this error.


travispotterBH:
@BHoMBot check required

pawelbaran (changes requested):
As a raw draft it looks good @travispotterBH 👍 good stuff in general. However, there is quite a bit to be improved in terms of code organisation:
- see the compliance check report by @BHoMBot 
- we try to avoid _Get_ prefix in `Query` method naming, assuming that it is self-evident that we want to get something by making a query - this is not a strong convention, but a general rule worth following if possible (_Is_ prefix is fine, though)
- `Checkout` looks like a `Modify` method to me (or `Compute`, but def not `Query`)
- warnings can stay private if that works for you, but you could also make them internal and move to Compute/Warnings

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,92 @@
...
+            else if (element.IsOwnedByCurrentUser())
+            {
+                Compute.ElementOwnedByCurrentUserWarning(element);
+            }
+
+            Compute.ElementOwnedByOtherUserWarning(element);
+
+        }
+
+        [Description("Modifies CheckoutStatus for each element selectd if element is not currently owned by the current user or others.")]
```
vietle-bh, line 61: There may be a typo here. I often use[ this VS plugin](https://github.com/EWSoftware/VSSpellChecker) to check my spelling 😉

travispotterBH:
@BHoMBot check required

Revit_Core_Engine/Modify/Checkout.cs
FraserGreenroyd: Make sure the plugin is set for British English 😉

vietle-bh: Good call Fraser! 😎 
image

travispotterBH:
@BHoMBot  check copyright-compliance

Revit_Core_Engine/Query/ElementsEditablePerCheckoutStatus.cs
```diff
@@ -0,0 +1,60 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+        [Description("Returns a sublist of elements, from the given list of elements, that are editable per the element's CheckoutStatus." +
+            "Return warnings option, if true, alerts the user of which elements are uneditable due to ownership by other users.")]
+        [Input("elements", "Revit elements.")]
+        [Input("returnWarnings", "If it is desired, a value of 'True' will raise warnings on which elements"
+            +"in the list are uneditable due to ownership by other users.")]
+        [Output("elementsEditablePerCheckoutStatus", "List of elements that are editable per the element's CheckoutStatus.")]
+        public static List<Element> ElementsEditablePerCheckoutStatus(this List<Element> elements, bool returnWarnings)
```
pawelbaran, line 46:
```suggestion
        public static List<Element> ElementsEditablePerCheckoutStatus(this List<Element> elements, bool recordWarnings)
```

Revit_Core_Engine/Query/IsOwnedByCurrentUser.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Input("element", "Revit element.")]
+        [Output("ownedByCurrentUser", "True if the input Revit element is owned by current user, otherwise false.")]
+        public static bool IsOwnedByCurrentUser(this Element element)
+        {
+            if (WorksharingUtils.GetCheckoutStatus(element.Document, element.Id) == CheckoutStatus.OwnedByCurrentUser)
+            {
+                return true;
+            }
+
+            return false;
```
pawelbaran, line 51:
```suggestion
            return element.OwnershipStatus() == CheckoutStatus.OwnedByCurrentUser;
```

Revit_Core_Engine/Query/IsOwnedByNone.cs
```diff
@@ -0,0 +1,55 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+        
+        [Description("Check if element is currently unowned.")]
+        [Input("element", "Revit element.")]
+        [Output("ownedByNone", "True if the input Revit element is not owned, otherwise false.")]
+        public static bool IsOwnedByNone(this Element element)
+        {
+            if (WorksharingUtils.GetCheckoutStatus(element.Document, element.Id) == CheckoutStatus.NotOwned)
```
pawelbaran, line 46: same as above

Revit_Core_Engine/Query/IsOwnedByOtherUser.cs
```diff
@@ -0,0 +1,55 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+       
+        [Description("Check if element is owned by another user.")]
+        [Input("element", "Revit element.")]
+        [Output("ownedByOtherUser", "True if the input Revit element is owned by another user, otherwise false.")]
+        public static bool IsOwnedByOtherUser(this Element element)
```
pawelbaran, line 44: same as above

Revit_Core_Engine/Query/OwnershipStatus.cs
```diff
@@ -0,0 +1,50 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+       
+        [Description("Returns element CheckoutStatus.")]
+        [Input("element", "Revit element.")]
+        [Output("checkoutStatus", "The worksharing ownership/CheckoutStatus of the element.")]
+        public static CheckoutStatus OwnershipStatus(this Element element)
```
pawelbaran, line 44: I would rename it to `CheckoutStatus` to keep consistent naming

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            Compute.ElementOwnedByOtherUserWarning(element);
+
+        }
+
+        [Description("Modifies CheckoutStatus for each element selected if element is not currently owned by the current user or others.")]
+        [Input("elements", "Revit elements.")]
+        public static void Checkout(this List<Element> elements)
+        {
```
pawelbaran, line 64: Here you are checking the same thing 3 times, which is far from optimal. Instead, you could do the following:
```suggestion
        {
            List<ElementId> toCheckout = new List<ElementId>();
            foreach (Element element in elements)
            {
                CheckoutStatus status = element.CheckoutStatus();
                // here go warnings depending on the status
                
                toCheckout.Add(element.Id);
            }
            
            // the line below will not work because you do not have one document value here - you can use Linq grouping to group by documents and call this per each doc
            WorksharingUtils.CheckoutElements(document, toCheckout);
        }
```

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,92 @@
...
+                Compute.ElementOwnedByCurrentUserWarning(element);
+            }
+
+            Compute.ElementOwnedByOtherUserWarning(element);
+
+        }
+
+        [Description("Modifies CheckoutStatus for each element selected if element is not currently owned by the current user or others.")]
+        [Input("elements", "Revit elements.")]
+        public static void Checkout(this List<Element> elements)
```
pawelbaran, line 63: The method should return `List<ElementId>`

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static partial class Modify
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+
+        [Description("Modifies element CheckoutStatus if element is not currently owned by the current user or others.")]
+        [Input("element", "Revit element.")]
+        public static void Checkout(this Element element)
```
pawelbaran, line 40: Do we ever need a checkout for a single element? It would be quite inefficient to call `CheckoutElements` per each single item. So I would opt for removing this method altogether if you do not have immediate use for it

Revit_Core_Engine/Query/ElementsOwnedByCurrentUser.cs
```diff
@@ -0,0 +1,50 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns a sublist of elements, from the given list of elements, that are owned by the current user.")]
+        [Input("elements", "Revit elements.")]
+        [Output("elementsOwnedByCurrentUser", "List of elements that are owned by the current user.")]
+        public static List<Element> ElementsOwnedByCurrentUser(this List<Element> elements)
+        {
+            return elements.Where(e => WorksharingUtils.GetCheckoutStatus(e.Document, e.Id) == CheckoutStatus.OwnedByCurrentUser).ToList();
```
pawelbaran, line 46: Same to other methods of this type
```suggestion
            return elements.Where(e => e.IsOwnedByCurrentUser()).ToList();
```

Revit_Core_Engine/Query/ElementsEditablePerCheckoutStatus.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (returnWarnings)
+            {
+                foreach (var element in elements.ElementsOwnedByOtherUsers())
+                {
+                    Compute.ElementOwnedByOtherUserWarning(element);
+                }
+            }
+
+            return elements.Where(e => WorksharingUtils.GetCheckoutStatus(e.Document, e.Id) != CheckoutStatus.OwnedByOtherUser).ToList();
```
pawelbaran, line 56:
```suggestion
            return elements.Where(e => !e.IsOwnedByOtherUser()).ToList();
```

Revit_Core_Engine/Compute/Warnings.cs
```diff
@@ -514,6 +514,26 @@ internal static void TransformNotImplementedWarning(this Element element)
...
+        /***************************************************/
+
+        internal static void ElementOwnedByCurrentUserWarning(Element element)
+        {
+            BH.Engine.Base.Compute.RecordWarning($"Revit object with ElementId: {element.Id} is owned by the current user.");
+        }
+
+        /***************************************************/
+
+        internal static void ElementOwnedByNoneWarning(Element element)
```
pawelbaran, line 532: This is never used, please remove

Revit_Core_Engine/Compute/Warnings.cs
```diff
@@ -514,6 +514,26 @@ internal static void TransformNotImplementedWarning(this Element element)
...
         /***************************************************/
+
+        internal static void ElementOwnedByOtherUserWarning(Element element)
+        {
+            BH.Engine.Base.Compute.RecordWarning($"Revit object could not be updated or modified due to it's CheckoutStatus. Revit ElementId: {element.Id} is owned by another user.");
+        }
+
+        /***************************************************/
+
+        internal static void ElementOwnedByCurrentUserWarning(Element element)
```
pawelbaran, line 525: I do not find it being a warning - a note at best. But would rather not record it as event altogether, assuming it is the default case.

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,92 @@
...
+            List<ElementId> elementsToCheckout = new List<ElementId>();
+
+            foreach (var element in elements.ElementsOwnedByOtherUsers())
+            {
+                Compute.ElementOwnedByOtherUserWarning(element);
+            }
+
+            foreach (var element in elements.ElementsOwnedByCurrentUser())
+            {
+                Compute.ElementOwnedByCurrentUserWarning(element);
```
pawelbaran, line 75: I would not record it as a warning here - what are you warning the user against? Being able to modify the element? 😉

pawelbaran (changes requested):

pawelbaran:
Please also try to make the PR title more human friendly (I would remove the issue number prefix)

Revit_Core_Engine/Modify/Checkout.cs
travispotterBH: Yeah, my thinking around this was that you can't checkout something that you've already checked out. So wanted to flag to the user that they already owned that element, so only x number of elements have been modified to be checked out.

pawelbaran: Right, then I would rather record note as irrelevant in most cases

Revit_Core_Engine/Query/ElementsEditablePerCheckoutStatus.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (recordWarnings)
+            {
+                foreach (var element in elements.ElementsOwnedByOtherUsers())
+                {
+                    Compute.ElementOwnedByOtherUserWarning(element);
+                }
+            }
+
+            return elements.Where(e => !e.IsOwnedByOtherUser()).ToList();
```
pawelbaran, line 56: Not sure how heavy it is to check the status of an element (if it requires talking to the server), so better avoid doing it twice:
```suggestion
            List<Element> result = new List<Element>();
            foreach(Element in elements)
            {
                // get checkout status, raise warning if needed, add to result if intended
            }
            return result;
```

Revit_Core_Engine/Query/ElementsOwnedByOtherUsers.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+        [Description("Returns a sublist of elements, from the given list of elements, that are owned by other users.")]
+        [Input("elements", "Revit elements.")]
+        [Output("elementsOwnedByUsers", "List of elements that are owned by other users.")]
+        public static List<Element> ElementsOwnedByOtherUsers(this List<Element> elements)
+        {
+            return elements.Where(e => e.CheckoutStatus()== Autodesk.Revit.DB.CheckoutStatus.OwnedByOtherUser).ToList();
```
pawelbaran, line 45:
```suggestion
            return elements.Where(e => e.IsOwnedByOtherUser()).ToList();
```

pawelbaran (changes requested):

travispotterBH:
@pawelbaran  I think we are getting to the point that we might be able to merge this. Though I am thinking about implementation of it. In any toolkit that attempts to modify Revit elements, we could have the developer of the methods remember to apply this in their workflow, though this to me can lead to issues in maintainability down the road. I'm considering that we should implement this on each of the `Modify` methods in this toolkit.  Such as [SetInsualtion](https://github.com/BHoM/Revit_Toolkit/blob/main/Revit_Core_Engine/Modify/SetInsulation.cs), [SetParameter](https://github.com/BHoM/Revit_Toolkit/blob/main/Revit_Core_Engine/Modify/SetParameter.cs), etc. Thoughts?

pawelbaran:
Agreed, we are getting close to getting this merged. Please do not bother about blanket-covering the entire toolkit for now, I think we can do it on adapter level, and if someone wants to use individual methods from this toolkit, one will need to remember to do the check explicitly.

travispotterBH:
@BHoMBot check required

Revit_Core_Engine/Modify/Checkout.cs
```diff
@@ -0,0 +1,70 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Modifies CheckoutStatus for each element selected if element is not currently owned by the current user or others.")]
+        [Input("elements", "Revit elements to query for its checkout status.")]
+        [Output("elementsCheckedOut", "List of elements checked out (ownership modified to be by current user) by this process.")]
+        public static List<ElementId> Checkout(this List<Element> elements)
+        {
+            Document document = elements.First().Document;
```
pawelbaran, line 42: You are missing a null check (if `elements == null`), could be also worth checking if all elements have same value of `Document` property (sounds like a `Linq` exercise)

Revit_Core_Engine/Query/ElementsEditablePerCheckoutStatus.cs
```diff
@@ -0,0 +1,64 @@
...
+            "Return warnings option, if true, alerts the user of which elements are uneditable due to ownership by other users.")]
+        [Input("elements", "Revit elements to query for its checkout status.")]
+        [Input("recordWarnings", "If it is desired, a value of 'True' will raise warnings on which elements"
+            +"in the list are uneditable due to ownership by other users.")]
+        [Output("elementsEditablePerCheckoutStatus", "List of elements that are editable per the element's CheckoutStatus.")]
+        public static List<Element> ElementsEditablePerCheckoutStatus(this List<Element> elements, bool recordWarnings)
+        {
+            List<Element> elementsOwnedByOthers = elements.ElementsOwnedByOtherUsers().ToList();
+
+           List<Element> elementsNotOwnedByOthers = elements.Except(elementsOwnedByOthers).ToList();
```
pawelbaran, line 50: Neat 👍

Revit_Core_Engine/Query/CheckoutStatus.cs
```diff
@@ -0,0 +1,50 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+       
+        [Description("Returns element CheckoutStatus.")]
+        [Input("element", "Revit element to query for its checkout status.")]
+        [Output("checkoutStatus", "The worksharing ownership/CheckoutStatus of the element.")]
+        public static CheckoutStatus CheckoutStatus(this Element element)
+        {
+            return WorksharingUtils.GetCheckoutStatus(element.Document, element.Id);
```
pawelbaran, line 46: We should be adding null checks on inputs where possible, so to improve this I would suggest
```suggestion
            if (element != null
                return WorksharingUtils.GetCheckoutStatus(element.Document, element.Id);
            else
                return null;
```

Revit_Core_Engine/Query/ElementsOwnedByCurrentUser.cs
```diff
@@ -0,0 +1,50 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns a sublist of elements, from the given list of elements, that are owned by the current user.")]
+        [Input("elements", "Revit elements to query for its checkout status.")]
+        [Output("elementsOwnedByCurrentUser", "List of elements that are owned by the current user.")]
+        public static List<Element> ElementsOwnedByCurrentUser(this List<Element> elements)
+        {
+            return elements.Where(e => e.IsOwnedByCurrentUser()).ToList();
```
pawelbaran, line 46:
```suggestion
            return elements?.Where(e => e.IsOwnedByCurrentUser()).ToList();
```

pawelbaran (changes requested):
Last minor suggestions made, only about input checks (few suggestions included, I bet you will be able to do similar in other methods). But we are definitely getting there 👍

travispotterBH:
New PR for branch name compliance: https://github.com/BHoM/Revit_Toolkit/pull/1248

PR #1247, opened 2022-08-30
Revit_Engine: Improve error message when diffing with duplicate elementIds

### LINKED ISSUE #1246: Revit_Engine: Improve messaging for RevitDiffing when duplicate ElementIds are detected
#### Description:

The current message for when objects with duplicate ElementIds have been provided for the RevitDiffing is a bit to hard to understand, given it is raised inside the main Diffing_Engine and need to be kept generic.

The objects should be checked in the RevitDiffing method and a more Revit specific message given, as well as guidance what can be done.

IsakNaslundBh:

Closes #1246 

Improves the error message given when duplicate ids is detected. Gives further guidance to use physical elements rather than structural if structural elements are detected.

Changing the Diffing method called to one one step deeper as ids are already extracted.

### Test files

(from @alelom) added a [test for this PR in DiffingTests_Prototypes](https://github.com/BHoM/DiffingTests_Prototypes/commit/4fd6beb6341649b0a350acec1f9fc8f29fd2d00a#diff-12ea9316c7af69edaf879938250d10c04a67762da06fdfc224959d42eb8e8c80). Run all tests in that solution to verify this PR.

### DIFF
--- a/Revit_Engine/Compute/RevitDiffing.cs
+++ b/Revit_Engine/Compute/RevitDiffing.cs
@@ -202,7 +202,7 @@ namespace BH.Engine.Adapters.Revit
                 return null;
             }
 
-            //Compute the diffing through DIffWIthCUstomIds making use of the ids extracted from the objects
+            //Compute the diffing through DiffWithCustomIds making use of the ids extracted from the objects
             Diff revitDiff = BH.Engine.Diffing.Compute.DiffWithCustomIds(pastObjects.ToList(), pastIds, followingObjects.ToList(), followingIds, diffConfigClone);
             return revitDiff;
         }
--- a/Revit_Engine/Compute/RevitDiffing.cs
+++ b/Revit_Engine/Compute/RevitDiffing.cs
@@ -191,7 +191,7 @@ namespace BH.Engine.Adapters.Revit
             {
                 List<string> messageSubjects = new List<string>();
                 if (dupsInPastIds) messageSubjects.Add(nameof(pastObjects));
-                if (dupsInFollIds) messageSubjects.Add(nameof(followingIds));
+                if (dupsInFollIds) messageSubjects.Add(nameof(followingObjects));
                 string message = $"Some of the {string.Join(" and ", messageSubjects)} contain duplicate {revitIdName}s. ";
 
                 if((dupsInPastIds && pastObjects.Any(x => x.GetType().Namespace.Contains("Structure"))) || 
--- a/Revit_Engine/Compute/RevitDiffing.cs
+++ b/Revit_Engine/Compute/RevitDiffing.cs
@@ -155,24 +155,59 @@ namespace BH.Engine.Adapters.Revit
             DiffingConfig diffConfigClone = GetRevitDiffingConfig(rcc);
 
             // Check if input objects all have RevitIdentifiers assigned.
-            IEnumerable<IBHoMObject> revitBHoMObjects_past = pastObjects.OfType<IBHoMObject>().Where(obj => obj.GetRevitIdentifiers() != null);
-            IEnumerable<IBHoMObject> revitBHoMObjects_following = followingObjects.OfType<IBHoMObject>().Where(obj => obj.GetRevitIdentifiers() != null);
+            IEnumerable<RevitIdentifiers> pastIdFragments = pastObjects.OfType<IBHoMObject>().Select(obj => obj.GetRevitIdentifiers()).Where(x => x != null);
+            IEnumerable<RevitIdentifiers> followingIdFragments = followingObjects.OfType<IBHoMObject>().Select(obj => obj.GetRevitIdentifiers()).Where(x => x != null);
 
-            if (revitBHoMObjects_past.Count() != pastObjects.Count())
+            if (pastIdFragments.Count() != pastObjects.Count())
             {
                 BH.Engine.Base.Compute.RecordError($"Some of the {nameof(pastObjects)} do not have a {nameof(RevitIdentifiers)} fragment attached.");
                 return null;
             }
 
-            if (revitBHoMObjects_following.Count() != followingObjects.Count())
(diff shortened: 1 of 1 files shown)


Revit_Engine/Compute/RevitDiffing.cs
```diff
@@ -155,24 +155,59 @@ private static Diff Diffing(IEnumerable<object> pastObjects, IEnumerable<object>
             DiffingConfig diffConfigClone = GetRevitDiffingConfig(rcc);
 
             // Check if input objects all have RevitIdentifiers assigned.
-            IEnumerable<IBHoMObject> revitBHoMObjects_past = pastObjects.OfType<IBHoMObject>().Where(obj => obj.GetRevitIdentifiers() != null);
-            IEnumerable<IBHoMObject> revitBHoMObjects_following = followingObjects.OfType<IBHoMObject>().Where(obj => obj.GetRevitIdentifiers() != null);
+            IEnumerable<RevitIdentifiers> pastIdFragments = pastObjects.OfType<IBHoMObject>().Select(obj => obj.GetRevitIdentifiers()).Where(x => x != null);
```
pawelbaran, line 158: Do you need to change this section of the method? It works, I would rather not touch it if possible, mainly to avoid the risk of breaking something

IsakNaslundBh: technically do not _need_ to change it, but I do need the line I added to be able to get the ids out. The previous collections were only used for a count check, and I cannot see any scenario where the count of the new collections would be different than the new ones. The only difference is the return type, which is used now, but was not before.

Happy to chat through if you think there are cases where the count could be different, but I honestly cant see it being possible, even if I tried to force it.

alelom: tl;dr: I think this is actually fine.

Longer reasoning: I think that I originally configured the method like this (getting the revit objects having a revit identifier, rather than directly getting the revit identifiers) because I thought that one could fall back on other Diffing methods not requiring a fragment Id attached, for the objects not owning one. However, as the solution took form, I formalized the `RevitDiffing()` method as only working with Fragments anyway, leaving the responsibility of non-fragment based diffing methodologies to the base Diffing_Engine.

Revit_Engine/Compute/RevitDiffing.cs
```diff
@@ -155,24 +155,59 @@ private static Diff Diffing(IEnumerable<object> pastObjects, IEnumerable<object>
             DiffingConfig diffConfigClone = GetRevitDiffingConfig(rcc);
 
             // Check if input objects all have RevitIdentifiers assigned.
```
alelom, line 157:
```suggestion
            // Get the past and following RevitIdentifiers fragments.
```

Revit_Engine/Compute/RevitDiffing.cs
```diff
@@ -155,24 +155,59 @@ private static Diff Diffing(IEnumerable<object> pastObjects, IEnumerable<object>
...
+            if (followingIds.Count != followingIds.Distinct().Count())
+            {
+                string message = $"Some of the {nameof(followingIds)} contain duplicate {revitIdName}s.";
+
+                if (pastObjects.Any(x => x.GetType().Namespace.Contains("Structure")))
+                    message += " If trying to diff structural objects, try pulling using the Phsyical discipline rather than Structural discipline, as models containing disjointed floors and walls and/or curved beams lead to one Revit element being converted into multiple structural BHoM elements.";
+
+                BH.Engine.Base.Compute.RecordError(message);
+                return null;
+            }
```
alelom, line 207: To avoid repeating the message contents, I thought we could rewrite this part as this:
(also there was an error on line 202: the check should've been with `followingObjects`)
```suggestion
             //Check for duplicate ids
            bool dupsInPastIds = pastIds.Count != pastIds.Distinct().Count();
            bool dupsInFollIds = followingIds.Count != followingIds.Distinct().Count();

            if (dupsInPastIds || dupsInFollIds)
            {
                List<string> messageSubjects = new List<string>();
                if (dupsInPastIds) messageSubjects.Add(nameof(pastObjects));
                if (dupsInFollIds) messageSubjects.Add(nameof(followingIds));
                string message = $"Some of the {string.Join(" and ", messageSubjects)} contain duplicate {revitIdName}s. ";

                if((dupsInPastIds && pastObjects.Any(x => x.GetType().Namespace.Contains("Structure"))) || 
                    (dupsInFollIds && followingObjects.Any(x => x.GetType().Namespace.Contains("Structure"))))
                    message += "\nIf trying to diff structural objects, try pulling using the Phsyical discipline rather than Structural discipline, as models containing disjointed floors and walls and/or curved beams lead to one Revit element being converted into multiple structural BHoM elements.";

                BH.Engine.Base.Compute.RecordError(message);
                return null;
            }
```

alelom (changes requested):
Some optional changes requested - happy to chat.

pawelbaran:
Thank you @alelom for the in-depth review. I think there is nothing left for me 😉 so happy to get it merged as soon as @alelom is happy with the changes 👍

IsakNaslundBh:
Thanks @alelom !

Have commited your suggested change with a minor fix and fixed a typo in the comments :)

alelom:
I've added a added a [test for this PR in DiffingTests_Prototypes](https://github.com/BHoM/DiffingTests_Prototypes/commit/4fd6beb6341649b0a350acec1f9fc8f29fd2d00a#diff-12ea9316c7af69edaf879938250d10c04a67762da06fdfc224959d42eb8e8c80).

IsakNaslundBh:
@alelom I have run through all tests, with all passing. 

If you are happy, would you mind approving so that we can get this merged? :)

alelom (approved):
Approved based on changes, discussion and tests.

IsakNaslundBh:
@BHoMBot check required

pawelbaran:
@BHoMBot check ready-to-merge

PR #1260, opened 2022-09-23
Duct Elements Convert Method Modified to Consider Disciplines

### LINKED ISSUE #1214: Revit_Core_Engine: Diffing MEP elements failed due to location curve split
While solving an unrelated issue, I noticed that diffing MEP elements such as ducts would always fail. This is because a duct can be exported as multiple objects with the same Guid to handle the case below.

https://github.com/BHoM/Revit_Toolkit/blob/15cfcd06a0f99ac02691f29ff175932ec595cf93/Revit_Core_Engine/Query/LocationCurve.cs#L163-L169

In this scenario, a diff will return null at this check...
https://github.com/BHoM/BHoM_Engine/blob/bdde55560efd5460930a858b4a5c1c9392f893e0/Diffing_Engine/Compute/DiffWithCustomIds.cs#L110-L111

... which means Revit isn't too happy:
image

Should we make it so all diffing called from Revit have `diffingConfig.AllowDuplicateIds` set to True?
Or, will we never diff MEP elements so this shouldn't be an issue?

travispotterBH:
I did not know the best place for this PR, or what was the best name of the branch, so I am prepared to remake this after comments.

Closes #1214 
Closes https://github.com/BHoM/BHoM/issues/1403

Attempts to address issue of duct objects being converted from Revit only as their analytical representation, which splits the ducts on all connected connectors (mostly only an issue on ducts with tap style fittings)

Try to Push to JSON a selected duct run with runs connected by taps and elbows like below: 
MicrosoftTeams-image (10)

export the same duct run as a physical discipline and as an environmental discipline. 

On the case of runs with tap fittings:
Physical Discipline should export just as the single, not separated by midpoint, Revit Element.
Environmental Discipline should export as the separated by midpoint element and should have multiple bhomobjects of type Duct of the same Revit Element Id. 

Runs without tap fittings should export similar to the Physical discipline would. 

MicrosoftTeams-image (9)

This was achieved by changing the `FromRevit` `Convert` Method for ducts called `DuctFromRevit` to take in the discipline provided, and that is passed into the `Query.LocationCurveMEP` method which then will either split or not split the duct line based on the discipline given. 

This method of doing this required edits to the `FromRevit` Convert method that handles `Element` types where it was required the new Invocation of the `convertMethod` would be able to handle additional parameters, namely the `Discipline` type. This further required a change to `Query.AllConvertMethods` to check for methods that have between three and four parameters instead of what was previous set to three.

@pawelbaran I recognize that this may not be the most optimal solution, and I welcome your comments/suggestions/assistance.

@kayleighhoude can you give some indication on how well this suits your needs. 

### Test files

Revit file: Any revit file with some ducts and fittings connected in a duct run
Expected Results:
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/Revit_Core_Engine-%231260-DuctElementsCreateMethodModifiedtoConsiderDisciplines?csf=1&web=1&e=3Gpp5P

### DIFF
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/Duct.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/Duct.cs
@@ -24,6 +24,7 @@ using Autodesk.Revit.DB;
 using BH.Engine.Adapters.Revit;
 using BH.Engine.Geometry;
 using BH.oM.Adapters.Revit;
+using BH.oM.Adapters.Revit.Enums;
 using BH.oM.Adapters.Revit.Settings;
 using BH.oM.Base;
 using BH.oM.Base.Attributes;
@@ -44,7 +45,7 @@ namespace BH.Revit.Engine.Core
         [Input("settings", "Revit adapter settings.")]
         [Input("refObjects", "A collection of objects processed in the current adapter action, stored to avoid processing the same object more than once.")]
         [Output("ducts", "List of BHoM duct objects converted from a Revit duct elements.")]
-        public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.Mechanical.Duct revitDuct, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null)
+        public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.Mechanical.Duct revitDuct, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null, Discipline discipline = Discipline.Physical)
         {
             settings = settings.DefaultIfNull();
 
@@ -59,7 +60,7 @@ namespace BH.Revit.Engine.Core
                 bhomDucts = new List<BH.oM.MEP.System.Duct>();
             }
             
-            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitDuct, settings);
+            List<BH.oM.Geometry.Line> queried = Query.LocationCurveMEP(revitDuct, settings, discipline);
             
             // Flow rate
             double flowRate = revitDuct.LookupParameterDouble(BuiltInParameter.RBS_DUCT_FLOW_PARAM); 
--- a/Revit_Core_Engine/Query/LocationCurve.cs
+++ b/Revit_Core_Engine/Query/LocationCurve.cs
@@ -32,6 +32,7 @@ using System.ComponentModel;
 using System.Linq;
 using BH.oM.Base.Attributes;
 using BH.oM.Base;
+using BH.oM.Adapters.Revit.Enums;
 
 namespace BH.Revit.Engine.Core
 {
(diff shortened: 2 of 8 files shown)


Revit_Core_Engine/Convert/MEP/FromRevit/Duct.cs
```diff
@@ -44,7 +45,7 @@ public static partial class Convert
         [Input("settings", "Revit adapter settings.")]
         [Input("refObjects", "A collection of objects processed in the current adapter action, stored to avoid processing the same object more than once.")]
         [Output("ducts", "List of BHoM duct objects converted from a Revit duct elements.")]
-        public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.Mechanical.Duct revitDuct, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null)
+        public static List<BH.oM.MEP.System.Duct> DuctFromRevit(this Autodesk.Revit.DB.Mechanical.Duct revitDuct, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null, Discipline discipline = Discipline.Physical)
```
pawelbaran, line 48: `Discipline` method parameter should not be added to this method - if you need it, please make this method `private` and then expose it via one-line public converts for physical and MEP disciplines that would call it.

Revit_Core_Engine/Query/LocationCurve.cs
```diff
@@ -160,7 +161,7 @@ public static BH.oM.Geometry.Line LocationCurveColumn(this FamilyInstance family
         [Input("mepCurve", "Revit MEPCurve to be queried.")]
         [Input("settings", "Optional, the RevitSettings to be used.")]
         [Output("locationCurveMEP", "BHoM lines list queried from the MEPCurve.")]
-        public static List<BH.oM.Geometry.Line> LocationCurveMEP(this MEPCurve mepCurve, RevitSettings settings = null)
+        public static List<BH.oM.Geometry.Line> LocationCurveMEP(this MEPCurve mepCurve, RevitSettings settings = null, Discipline discipline = Discipline.Physical)
```
pawelbaran, line 164: This is fine, there is no limitations on the `Query` methods.

pawelbaran (changes requested):
I am afraid you are jumping too far ahead with this PR @travispotterBH. The current system around the Revit converts is based on the concept of having one convert per object type. This means you cannot have 2 types of converts to MEP ducts depending on discipline input (see the code comments).

Instead, you would need to start from adding MEP objects to the `Physical` namespace (Physical_oM) to then start supporting them in Revit_Toolkit. I hope this makes sense - even if taking a shortcut sounds tempting in short term, I am pretty sure we would need to pay double price once we realise the amount of extra effort required to make it work plus potentially refactor in long term.

travispotterBH:
Will not planned to be resolved in this sprint.

PR #1281, opened 2022-11-24
Revit_Core_Engine: pull of windows from link fixed

### LINKED ISSUE #1280: Windows Sometimes Fail When Pulling from Link
#### Description:
Getting repeatable failure of window pulls from a link resulting from an issue here:
WindowLinkPullBug

#### Steps to reproduce:
Pull windows from a linked file with multiple walls with windows. Some entire walls of windows will pull, while some will all fail on a per wall basis.

#### Expected behaviour:
Windows pulling as per typical window pulls.

pawelbaran:

Closes #1280

### Test files

See https://github.com/BHoM/Revit_Toolkit/issues/1280

### Additional comments

From the two issues raised in the original issue:
Null reference was an easy fix, occurred when the solid was entirely outside of the half space, which makes perfect sense.

Generating ids error originates from an edge case where the element had its geometry defined both as instance geometry and element geometry (from what I understand, instance geometry is the geometry created based on the family definition, while the element geometry is what it actually is in the model - see [Revit API docs](https://www.revitapidocs.com/2018.2/fe25b14f-5866-ca0f-a660-c157484c3a56.htm) for more details):
- usually walls have only element geometry, for some weird reason some of them also have instance one:
    image
        ...from what I understand, instance geometry should be ignored in this case because it does not really exist in the model
- `FamilyInstance` _should_ only have instance geometry, but who knows what will Revit come up with...
- I have found elements that have both element and instance geometry (see railing below), but apparently the element geometry is invalid then
    image
    
...so I drew a conclusion that the procedure should be as follows:
- if the element has a valid element geometry, return only this geometry
- otherwise return instance geometry

I made the above assumptions after longer investigation, but I cannot fully guarantee that it is valid for all cases. Therefore, I strongly recommend wider testing using the standard testing procedures before this gets merged @enarhi.

### DIFF
--- a/Revit_Core_Engine/Query/GeometryPrimitives.cs
+++ b/Revit_Core_Engine/Query/GeometryPrimitives.cs
@@ -45,7 +45,8 @@ namespace BH.Revit.Engine.Core
             if (geometryElement == null)
                 return null;
 
-            List<GeometryObject> result = new List<GeometryObject>();
+            List<GeometryObject> geometry = new List<GeometryObject>();
+            List<GeometryObject> instanceGeometry = new List<GeometryObject>();
             foreach (GeometryObject geometryObject in geometryElement)
             {
                 if (geometryObject is GeometryInstance)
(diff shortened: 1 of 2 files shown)


pawelbaran:
@BHoMBot check required

enarhi (approved):
LGTM. Tested with the original test file and the pull bug is fixed. Also tested Pulling Physical Panel Locations and Pulling FacadeEElements from the Revit test procedures, and both worked as expected. @pawelbaran please flag if other testprocedures may be required based on your changes, otherwise this one is good to go from my end.

pawelbaran:
@BHoMBot check ready-to-merge

PR #1289, opened 2022-12-13
Revit_Core_Engine: Handle null exception on Nurbs curves in beams

### LINKED ISSUE #1288: Null exception on Nurbs
#### Description:

This line:
https://github.com/BHoM/Revit_Toolkit/blob/c27a9a615f07ee9eff9870be94b51360a975b22b/Revit_Core_Engine/Convert/Structure/FromRevit/Bars.cs#L136
gives a null exception `locationCurve.ICollapseToPolyline(Math.PI / 12) = null` when `locationCurve` is a Nurbs.

#### Steps to reproduce:

Simply pull all beams from the test model below.

#### Expected behaviour:

Not sure if we fix it now in Revit_Toolkit or wait for Bhom_Engine to support nurbs.

#### Test file(s):

On SharePoint

vietle-bh:

Closes #1288 

While waiting for Bhom_Engine to add support for Nurbs, at least let's handle the exception on the Revit side by converting nurbs curves to polylines.

image

### Test files

On [SharePoint](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/Revit_Core_Engine-%231288-NullException?csf=1&web=1&e=giCgaQ)

### Changelog

- Convert Nurbs curves to polylines to handle a null exception

### DIFF
--- a/Revit_Core_Engine/Convert/Structure/FromRevit/Bars.cs
+++ b/Revit_Core_Engine/Convert/Structure/FromRevit/Bars.cs
@@ -28,6 +28,7 @@ using BH.Engine.Geometry;
 using BH.oM.Adapters.Revit.Settings;
 using BH.oM.Base;
 using BH.oM.Base.Attributes;
+using BH.oM.Geometry;
 using BH.oM.Spatial.ShapeProfiles;
 using BH.oM.Structure.Elements;
 using BH.oM.Structure.MaterialFragments;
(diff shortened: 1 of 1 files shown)


vietle-bh:
@BHoMBot check versioning
@BHoMBot check installer

pawelbaran:
Thanks @vietle-bh for raising this. However, this one was left on purpose after a discussion with @IsakNaslundBh. There are 2 reasons for it:
- we are trying to avoid incidental checks if a curve is nurbs inside the code for maintenance reasons - it is virtually impossible to keep track of this sort of checks, e.g. when support for nurbs curves gets added in BHoM_Engine (which actually is slowly happening thanks to @IsakNaslundBh's hobby 😉), then we would need to parse the entire code base in search for this sort of ifs, which is error prone and not very efficient
- curve tessellation may generate curves that will not make sense from the engineering perspective, so we would rather not make this decision for the user - panels are a slightly different case because it is harder to draw them by hand (especially when they have funky shapes) and they usually require manual tweaks anyways

I hope this makes sense, @IsakNaslundBh please feel free to correct me if I missed something or you disagree

vietle-bh:
> I hope this makes sense, @IsakNaslundBh please feel free to correct me if I missed something or you disagree

Sure, I'll close this PR. How about [this other one](https://github.com/BHoM/Revit_Toolkit/pull/1291)?

pawelbaran:
> Sure, I'll close this PR. How about [this other one](https://github.com/BHoM/Revit_Toolkit/pull/1291)?

Responded 👍

PR #1294, opened 2022-12-21
BHoM Version Button Added

### LINKED ISSUE #1296: Add BHoM Button Version
#### Description:

Add Button to UI to allow users to check which version they have installed.

travispotterBH:

Closes #1296 

Users can now check their current BHoM installation version.

### Changelog

- Currently installed BHoM version button added.

### DIFF
--- a/Revit_Core_Adapter/Listener/Buttons/BHoMVersion.cs
+++ b/Revit_Core_Adapter/Listener/Buttons/BHoMVersion.cs
@@ -1,0 +1,47 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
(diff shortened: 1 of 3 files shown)


pawelbaran (commented):
This looks good to me and I have no objections, however I would like @FraserGreenroyd to give the final green light as this PR crosses with both UI as well as CI/CD

FraserGreenroyd (changes requested):
Rather than just displaying the BHoM version, I think this would serve well as being an `About` button which contains some key information users may find beneficial, such as links to the wiki documentation, etc.

travispotterBH:
> Rather than just displaying the BHoM version, I think this would serve well as being an `About` button which contains some key information users may find beneficial, such as links to the wiki documentation, etc.
Current Info tab includes link to [Github wiki](https://github.com/BHoM/documentation/wiki) and [bhom.xyz](https://bhom.xyz/)
image

This PR adds: 
image
image


I believe the pre-existing information buttons, along with the addition will satisfy the request for links to wiki documentation.

FraserGreenroyd:
But then why hide the version behind a button at all if that's all we're going to show? An about button can be expanded upon further to display installation date, it could show a more specific version (such as the alpha version if using an alpha over a beta).

If that link to the wiki is truly going to the BHoM Wiki (and not the Revit Wiki - which is the wiki that would be additionally pertinent on Revit Toolkit) then the link also needs updating to the new documentation link - http://bhom.xyz/documentation

travispotterBH:
Roadmapped: https://github.com/BHoM/BHoM_Engine/issues/2971

pawelbaran:
Sounds like there is a plan. I do not have a strong opinion here, so happy for you guys to lead, but please nudge me in case you need any support 👍 Meanwhile I will keep on lurking 🪱

FraserGreenroyd:
Discussed with @travispotterBH offline, going to close this for now in favour of closing out the roadmapped issue https://github.com/BHoM/BHoM_Engine/issues/2971 to produce a good version of this. If that takes too long though then we will come back to this work as a default for beta - so branch not being deleted at this stage.

PR #1313, opened 2023-02-17
Add ViewPlan, Sheet, Viewport and Drawing Area methods

### LINKED ISSUE #1314: Add ViewPlan, Ceiling Plan, Sheet, Place View On Sheet, Drawing Area methods
#### Description:

As in the title.

michal-pekacki:

Closes #1314 

Methods needed for generating views and sheets.

### DIFF
--- a/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
+++ b/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
@@ -21,6 +21,7 @@
  */
 
 using Autodesk.Revit.DB;
+using Autodesk.Revit.UI;
 using BH.oM.Base.Attributes;
 using System.ComponentModel;
 
@@ -43,6 +44,13 @@ namespace BH.Revit.Engine.Core
         public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = ViewportRotation.None)
         {
             Viewport viewPort = Viewport.Create(document, sheet.Id, view.Id, viewportPlacementPoint);
+
+            if (viewPort == null)
+            {
+                BH.Engine.Base.Compute.RecordWarning($"Could not create Viewport of the view '{view.Name}'. Please check if view isn't empty.");
+                return viewPort;
+            }
+
             viewPort.ChangeTypeId(viewportTypeId);
             viewPort.Rotation = viewportRotation;
 
--- a/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
+++ b/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
@@ -21,12 +21,8 @@
  */
 
 using Autodesk.Revit.DB;
-using BH.Engine.Geometry;
 using BH.oM.Base.Attributes;
-using BH.oM.Geometry;
-using System.Collections.Generic;
 using System.ComponentModel;
-using System.Linq;
 
 namespace BH.Revit.Engine.Core
 {
@@ -36,18 +32,24 @@ namespace BH.Revit.Engine.Core
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Executes initial filtering of elements to rule out closed worksets. Returns a collection of ElementIds or a null to be passed on to IElementIds method) for further filtering.")]
-        [Input("document", "Revit document to be processed.")]
-        [Output("prefilter", "Collection of all ElementIds that exist in the open worksets of the input document.")]
+        [Description("Places view on sheet in specified position and rotation using the specified viewport type.")]
+        [Input("document", "Revit current document to be processed.")]
+        [Input("sheet", "Sheet to place the view on.")]
+        [Input("view", "View to be placed on sheet.")]
+        [Input("viewportTypeId", "Id of the viewport to be used.")]
+        [Input("viewportPlacementPoint", "Placement point of the viewport on the sheet.")]
+        [Input("viewportRotation", "Rotation property of the viewport. None as default.")]
+        [Output("viewPort", "The new viewport.")]
         public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = 0)
         {
-
             Viewport viewPort = Viewport.Create(document, sheet.Id, view.Id, viewportPlacementPoint);
             viewPort.SetParameter(BuiltInParameter.ELEM_TYPE_PARAM, viewportTypeId);
             viewPort.Rotation = viewportRotation;
 
             document.Regenerate();
-            viewPort.SetBoxCenter(viewportPlacementPoint);
+
+            if (viewportPlacementPoint != null)
+                viewPort.SetBoxCenter(viewportPlacementPoint);
 
             return viewPort;
         }
--- a/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
+++ b/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
@@ -32,18 +32,18 @@ namespace BH.Revit.Engine.Core
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Places view on sheet in specified position and rotation using the specified viewport type.")]
-        [Input("document", "Revit current document to be processed.")]
+        [Description("Places view on sheet at specified position and rotation using the specified viewport type.")]
+        [Input("document", "The current Revit document to be processed.")]
         [Input("sheet", "Sheet to place the view on.")]
         [Input("view", "View to be placed on sheet.")]
         [Input("viewportTypeId", "Id of the viewport to be used.")]
         [Input("viewportPlacementPoint", "Placement point of the viewport on the sheet.")]
-        [Input("viewportRotation", "Rotation property of the viewport. None as default.")]
+        [Input("viewportRotation", "Rotation type of the viewport. The default is None.")]
         [Output("viewPort", "The new viewport.")]
-        public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = 0)
+        public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = ViewportRotation.None)
         {
             Viewport viewPort = Viewport.Create(document, sheet.Id, view.Id, viewportPlacementPoint);
-            viewPort.SetParameter(BuiltInParameter.ELEM_TYPE_PARAM, viewportTypeId);
+            viewPort.ChangeTypeId(viewportTypeId);
             viewPort.Rotation = viewportRotation;
 
             document.Regenerate();
--- a/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
+++ b/Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
@@ -1,0 +1,59 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2023, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using BH.Engine.Geometry;
+using BH.oM.Base.Attributes;
+using BH.oM.Geometry;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Executes initial filtering of elements to rule out closed worksets. Returns a collection of ElementIds or a null to be passed on to IElementIds method) for further filtering.")]
+        [Input("document", "Revit document to be processed.")]
+        [Output("prefilter", "Collection of all ElementIds that exist in the open worksets of the input document.")]
+        public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = 0)
+        {
+
+            Viewport viewPort = Viewport.Create(document, sheet.Id, view.Id, viewportPlacementPoint);
+            viewPort.SetParameter(BuiltInParameter.ELEM_TYPE_PARAM, viewportTypeId);
+            viewPort.Rotation = viewportRotation;
+
+            document.Regenerate();
+            viewPort.SetBoxCenter(viewportPlacementPoint);
+
+            return viewPort;
+        }
+
+        /***************************************************/
+
+    }
+}
+
--- a/Revit_Core_Engine/Create/Sheet.cs
+++ b/Revit_Core_Engine/Create/Sheet.cs
@@ -1,0 +1,79 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2023, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using BH.oM.Base.Attributes;
+using System.ComponentModel;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Input("document", "Revit current document to be processed.")]
+        [Input("sheetName", "Name of the new sheet.")]
+        [Input("sheetNumber", "Number of the new sheet.")]
+        [Input("viewTemplateId", "The Title Block Id to be applied in the sheet.")]
+        [Output("viewSheet", "The new sheet.")]
+        public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
+        {
+            ViewSheet result = null;
+
+            result = ViewSheet.Create(document, titleBlockId);
+
+            if (!string.IsNullOrEmpty(sheetName))
+            {
+                try
+                {
+                    result.Name = sheetName;
+                }
+                catch
+                {
+                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetName + "'." + " It has been named '" + result.Name + "' instead.");
+                }
+            }
+
+            if (!string.IsNullOrEmpty(sheetNumber))
+            {
+                try
+                {
+                    result.SheetNumber = sheetNumber;
+                }
+                catch
+                {
+                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet with number '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
+                }
+            }
+
+            return result;
+        }
+
+        /***************************************************/
+    }
+}
+
+
+
--- a/Revit_Core_Engine/Create/Sheet.cs
+++ b/Revit_Core_Engine/Create/Sheet.cs
@@ -33,16 +33,14 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
-        [Input("document", "Revit current document to be processed.")]
+        [Input("document", "The current Revit document to be processed.")]
         [Input("sheetName", "Name of the new sheet.")]
         [Input("sheetNumber", "Number of the new sheet.")]
-        [Input("viewTemplateId", "The Title Block Id to be applied in the sheet.")]
+        [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
         [Output("viewSheet", "The new sheet.")]
         public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
         {
-            ViewSheet result = null;
-
-            result = ViewSheet.Create(document, titleBlockId);
+            ViewSheet result = ViewSheet.Create(document, titleBlockId);
 
             if (!string.IsNullOrEmpty(sheetName))
             {
@@ -50,7 +48,7 @@ namespace BH.Revit.Engine.Core
                 {
                     result.Name = sheetName;
                 }
-                catch
+                catch (Autodesk.Revit.Exceptions.ArgumentException)
                 {
                     BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetName + "'." + " It has been named '" + result.Name + "' instead.");
                 }
@@ -62,7 +60,7 @@ namespace BH.Revit.Engine.Core
                 {
                     result.SheetNumber = sheetNumber;
                 }
-                catch
+                catch (Autodesk.Revit.Exceptions.ArgumentException)
                 {
                     BH.Engine.Base.Compute.RecordWarning("There is already a sheet with number '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
                 }
--- a/Revit_Core_Engine/Create/Sheet.cs
+++ b/Revit_Core_Engine/Create/Sheet.cs
@@ -44,25 +44,31 @@ namespace BH.Revit.Engine.Core
 
             if (!string.IsNullOrEmpty(sheetName))
             {
-                try
-                {
-                    result.Name = sheetName;
-                }
-                catch (Autodesk.Revit.Exceptions.ArgumentException)
-                {
-                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetName + "'." + " It has been named '" + result.Name + "' instead.");
-                }
+                result.Name = sheetName;
             }
 
             if (!string.IsNullOrEmpty(sheetNumber))
             {
-                try
+                int number = 0;
+
+                while (true)
                 {
-                    result.SheetNumber = sheetNumber;
-                }
-                catch (Autodesk.Revit.Exceptions.ArgumentException)
-                {
-                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet with number '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
+                    try
+                    {
+                        if (number != 0)
+                        {
+                            result.SheetNumber = $"{sheetNumber} ({number})";
+                            BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
+                            break;
+                        }
+
+                        result.SheetNumber = sheetNumber;
+                        break;
+                    }
+                    catch
+                    {
+                        number++;
+                    }
                 }
             }
 
--- a/Revit_Core_Engine/Create/View/Sheet.cs
+++ b/Revit_Core_Engine/Create/View/Sheet.cs
@@ -34,7 +34,7 @@ namespace BH.Revit.Engine.Core
         /****               Public Methods              ****/
         /***************************************************/
 
-        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Description("Creates and returns a new Sheet in the current Revit file.")]
         [Input("document", "The current Revit document to be processed.")]
         [Input("sheetName", "Name of the new sheet.")]
         [Input("sheetNumber", "Number of the new sheet.")]
@@ -42,7 +42,7 @@ namespace BH.Revit.Engine.Core
         [Output("newSheet", "The new sheet.")]
         public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
         {
-            List<string> sheetNumbersInModel = new FilteredElementCollector(document).OfClass(typeof(ViewSheet)).WhereElementIsNotElementType().Select(x => (x as ViewSheet).SheetNumber).ToList();
+            List<string> sheetNumbersInModel = new FilteredElementCollector(document).OfClass(typeof(ViewSheet)).Cast<ViewSheet>().Select(x => x.SheetNumber).ToList();
             ViewSheet newSheet = ViewSheet.Create(document, titleBlockId);
 
             if (!string.IsNullOrEmpty(sheetName))
@@ -64,7 +64,7 @@ namespace BH.Revit.Engine.Core
                 newSheet.SheetNumber = uniqueNumber;
                 if (uniqueNumber != sheetNumber)
                 {
-                    BH.Engine.Base.Compute.RecordWarning($"There is already a sheet named '{sheetNumber}'. It has been named '{uniqueNumber}' instead.");
+                    BH.Engine.Base.Compute.RecordWarning($"Sheet named '{sheetNumber}' already exists in the document. Newly created has been named '{uniqueNumber}' instead.");
                 }
             }
 
--- a/Revit_Core_Engine/Create/View/Sheet.cs
+++ b/Revit_Core_Engine/Create/View/Sheet.cs
@@ -22,7 +22,9 @@
 
 using Autodesk.Revit.DB;
 using BH.oM.Base.Attributes;
+using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 
 namespace BH.Revit.Engine.Core
 {
@@ -37,42 +39,36 @@ namespace BH.Revit.Engine.Core
         [Input("sheetName", "Name of the new sheet.")]
         [Input("sheetNumber", "Number of the new sheet.")]
         [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
-        [Output("viewSheet", "The new sheet.")]
+        [Output("newSheet", "The new sheet.")]
         public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
         {
-            ViewSheet result = ViewSheet.Create(document, titleBlockId);
+            List<string> sheetNumbersInModel = new FilteredElementCollector(document).OfClass(typeof(ViewSheet)).WhereElementIsNotElementType().Select(x => (x as ViewSheet).SheetNumber).ToList();
+            ViewSheet newSheet = ViewSheet.Create(document, titleBlockId);
 
             if (!string.IsNullOrEmpty(sheetName))
             {
-                result.Name = sheetName;
+                newSheet.Name = sheetName;
             }
 
             if (!string.IsNullOrEmpty(sheetNumber))
             {
                 int number = 0;
+                string uniqueNumber = sheetNumber;
 
-                while (true)
+                while (sheetNumbersInModel.Contains(uniqueNumber))
                 {
-                    try
-                    {
-                        if (number != 0)
-                        {
-                            result.SheetNumber = $"{sheetNumber} ({number})";
-                            BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
-                            break;
-                        }
+                    number++;
+                    uniqueNumber = $"{sheetNumber} ({number})";
+                }
 
-                        result.SheetNumber = sheetNumber;
-                        break;
-                    }
-                    catch
-                    {
-                        number++;
-                    }
+                newSheet.SheetNumber = uniqueNumber;
+                if (uniqueNumber != sheetNumber)
+                {
+                    BH.Engine.Base.Compute.RecordWarning($"There is already a sheet named '{sheetNumber}'. It has been named '{uniqueNumber}' instead.");
                 }
             }
 
-            return result;
+            return newSheet;
         }
 
         /***************************************************/
--- a/Revit_Core_Engine/Create/View/Sheet.cs
+++ b/Revit_Core_Engine/Create/View/Sheet.cs
@@ -38,7 +38,7 @@ namespace BH.Revit.Engine.Core
         [Input("document", "The current Revit document to be processed.")]
         [Input("sheetName", "Name of the new sheet.")]
         [Input("sheetNumber", "Number of the new sheet.")]
-        [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
+        [Input("titleBlockId", "The Title Block Id to be applied to the sheet.")]
         [Output("newSheet", "The new sheet.")]
         public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
         {
--- a/Revit_Core_Engine/Create/View/ViewPlan.cs
+++ b/Revit_Core_Engine/Create/View/ViewPlan.cs
@@ -88,24 +88,7 @@ namespace BH.Revit.Engine.Core
             
             if (!string.IsNullOrEmpty(viewName))
             {
-                int number = 0;
-                string uniqueName = viewName;
-
-                while (uniqueName.IsExistingViewName(document))
-                {
-                    number++;
-                    uniqueName = $"{viewName} ({number})";
-                }
-
-#if (REVIT2018 || REVIT2019)
-                    newView.ViewName = uniqueName;
-#else
-                newView.Name = uniqueName;
-#endif
-                if (uniqueName != viewName)
-                {
-                    BH.Engine.Base.Compute.RecordWarning($"There is already a view named '{viewName}'. It has been named '{uniqueName}' instead.");
-                }
+                newView.SetViewName(viewName, document);
             }
             
             return newView;
@@ -157,24 +140,7 @@ namespace BH.Revit.Engine.Core
 
             if (!string.IsNullOrEmpty(viewName))
             {
-                int number = 0;
-                string uniqueName = viewName;
-
-                while (uniqueName.IsExistingViewName(document))
-                {
-                    number++;
-                    uniqueName = $"{viewName} ({number})";
-                }
-
-#if (REVIT2018 || REVIT2019)
-                    newView.ViewName = uniqueName;
-#else
-                newView.Name = uniqueName;
-#endif
-                if (uniqueName != viewName)
-                {
-                    BH.Engine.Base.Compute.RecordWarning($"There is already a view named '{viewName}'. It has been named '{uniqueName}' instead.");
-                }
+                newView.SetViewName(viewName, document);
             }
 
             if (scopeBoxId != null)
@@ -205,6 +171,31 @@ namespace BH.Revit.Engine.Core
 
         /***************************************************/
         /****              Private Methods              ****/
+        /***************************************************/
+
+        [Description("Set View Name to the given value. If the view name already exists in the model, a number suffix is added.")]
+        private static void SetViewName(this View view, string viewName, Document document)
+        {
+            int number = 0;
+            string uniqueName = viewName;
+
+            while (uniqueName.IsExistingViewName(document))
+            {
+                number++;
+                uniqueName = $"{viewName} ({number})";
+            }
+
+#if (REVIT2018 || REVIT2019)
+                    view.ViewName = uniqueName;
+#else
+            view.Name = uniqueName;
+#endif
+            if (uniqueName != viewName)
+            {
+                BH.Engine.Base.Compute.RecordWarning($"There is already a view named '{viewName}'. It has been named '{uniqueName}' instead.");
+            }
+        }
+
         /***************************************************/
 
         [Description("Check if given view name exists in the Revit model.")]
--- a/Revit_Core_Engine/Create/View/ViewPlan.cs
+++ b/Revit_Core_Engine/Create/View/ViewPlan.cs
@@ -37,7 +37,7 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         [PreviousVersion("6.1", "BH.Revit.Engine.Core.Create.ViewPlan(Autodesk.Revit.DB.Document, Autodesk.Revit.DB.Level, System.String, Autodesk.Revit.DB.ViewFamily, Autodesk.Revit.DB.CurveLoop, Autodesk.Revit.DB.ElementId, Autodesk.Revit.DB.ViewDetailLevel)")]
-        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Description("Creates and returns a new Plan view in the current Revit file.")]
         [Input("document", "Revit current document to be processed.")]
         [Input("level", "The level that the created Floor Plan refers to.")]
         [Input("viewName", "Name of the new view.")]
@@ -111,9 +111,9 @@ namespace BH.Revit.Engine.Core
 
         /***************************************************/
 
-        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Description("Creates and returns a new Plan view in the current Revit file.")]
         [Input("document", "The current Revit document to be processed.")]
-        [Input("level", "The level that the created Floor Plan refers to.")]
+        [Input("level", "The level that the created view refers to.")]
         [Input("viewName", "Name of the new view.")]
         [Input("viewFamily", "View Family of the View Type. The default is FloorPlan.")]
         [Input("scopeBoxId", "(Optional) A Scope Box Id to attempt to apply to the newly created view.")]
--- a/Revit_Core_Engine/Create/View/ViewPlan.cs
+++ b/Revit_Core_Engine/Create/View/ViewPlan.cs
@@ -43,12 +43,19 @@ namespace BH.Revit.Engine.Core
         [Input("cropBox", "Optional, the crop region to attempt to apply to the newly created view.")]
         [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
         [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]        
-        [Output("newView", "The new view.")]        
-        public static View ViewPlan(this Document document, Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
+        [Output("newView", "The new view.")]
+        [PreviousVersion("6.1", "BH.Revit.Engine.Core.Create.ViewPlan(Autodesk.Revit.DB.Document, Autodesk.Revit.DB.Level, System.String, Autodesk.Revit.DB.ViewFamily, Autodesk.Revit.DB.CurveLoop, Autodesk.Revit.DB.ElementId, Autodesk.Revit.DB.ViewDetailLevel)")]
+        public static View ViewPlan(this Document document, Level level, string viewName = null, ViewFamily viewFamily = ViewFamily.FloorPlan, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
         {
             View newView = null;
 
-            ViewFamilyType viewFamilyType = Query.ViewFamilyType(document, ViewFamily.FloorPlan);
+            if (!(viewFamily == ViewFamily.FloorPlan || viewFamily == ViewFamily.CeilingPlan || viewFamily == ViewFamily.AreaPlan || viewFamily == ViewFamily.StructuralPlan))
+            {
+                BH.Engine.Base.Compute.RecordWarning($"Could not create View of type '{viewFamily}'. It has to be a FloorPlan, CeilingPlan, AreaPlan, or StructuralPlan ViewType.");
+                return newView;
+            }
+
+            ViewFamilyType viewFamilyType = Query.ViewFamilyType(document, viewFamily);
 
             newView = Autodesk.Revit.DB.ViewPlan.Create(document, viewFamilyType.Id, level.Id);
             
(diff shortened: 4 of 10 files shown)


Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Places view on sheet in specified position and rotation using the specified viewport type.")]
```
vietle-bh, line 35:
```suggestion
        [Description("Places view on sheet at specified position and rotation using the specified viewport type.")]
```

Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
```diff
@@ -0,0 +1,61 @@
...
+        [Input("document", "Revit current document to be processed.")]
```
vietle-bh, line 36:
```suggestion
        [Input("document", "The current Revit document to be processed.")]
```

Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
```diff
@@ -0,0 +1,61 @@
...
+        [Input("sheet", "Sheet to place the view on.")]
+        [Input("view", "View to be placed on sheet.")]
+        [Input("viewportTypeId", "Id of the viewport to be used.")]
+        [Input("viewportPlacementPoint", "Placement point of the viewport on the sheet.")]
+        [Input("viewportRotation", "Rotation property of the viewport. None as default.")]
```
vietle-bh, line 41:
```suggestion
        [Input("viewportRotation", "Rotation type of the viewport. The default is None.")]
```

Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
```diff
@@ -0,0 +1,61 @@
...
+        [Output("viewPort", "The new viewport.")]
+        public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = 0)
```
vietle-bh, line 43:
```suggestion
        public static Viewport PlaceViewOnSheet (this Document document, ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = ViewportRotation.None)
```

Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            Viewport viewPort = Viewport.Create(document, sheet.Id, view.Id, viewportPlacementPoint);
+            viewPort.SetParameter(BuiltInParameter.ELEM_TYPE_PARAM, viewportTypeId);
```
vietle-bh, line 46: Would it be cleaner to use `Element.ChangeTypeId(typeId)`?

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Input("document", "Revit current document to be processed.")]
```
vietle-bh, line 36:
```suggestion
        [Input("document", "The current Revit document to be processed.")]
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Input("sheetName", "Name of the new sheet.")]
+        [Input("sheetNumber", "Number of the new sheet.")]
+        [Input("viewTemplateId", "The Title Block Id to be applied in the sheet.")]
```
vietle-bh, line 39:
```suggestion
        [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Output("viewSheet", "The new sheet.")]
+        public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
+        {
+            ViewSheet result = null;
+
+            result = ViewSheet.Create(document, titleBlockId);
```
vietle-bh, line 45:
```suggestion
            ViewSheet result = ViewSheet.Create(document, titleBlockId);
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (!string.IsNullOrEmpty(sheetName))
+            {
+                try
+                {
+                    result.Name = sheetName;
+                }
+                catch
```
vietle-bh, line 53:
```suggestion
                catch (Autodesk.Revit.Exceptions.ArgumentException)
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet named '" + sheetName + "'." + " It has been named '" + result.Name + "' instead.");
```
vietle-bh, line 55: Instead of using the new name Revit auto-generates, should we append (1)/(2)/(3)/etc. to the name the user has specified? 
If so, a while loop to increment this number until the new name is unique will do. The user can then rename the sheet as required much quicker.

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (!string.IsNullOrEmpty(sheetNumber))
+            {
+                try
+                {
+                    result.SheetNumber = sheetNumber;
+                }
+                catch
+                {
+                    BH.Engine.Base.Compute.RecordWarning("There is already a sheet with number '" + sheetNumber + "'." + " It has been named '" + result.SheetNumber + "' instead.");
```
vietle-bh, line 67: Instead of using the new number Revit auto-generates, should we append (1)/(2)/(3)/etc. to the number the user has specified? 
If so, a while loop to increment this number until the new name is unique will do. The user can then renumber the sheet as required much quicker.

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+            }
+
+            if (!string.IsNullOrEmpty(sheetNumber))
+            {
+                try
+                {
+                    result.SheetNumber = sheetNumber;
+                }
+                catch
```
vietle-bh, line 65:
```suggestion
                catch (Autodesk.Revit.Exceptions.ArgumentException)
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
         }
 
         /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Input("document", "Revit current document to be processed.")]
```
vietle-bh, line 102:
```suggestion
        [Input("document", "The current Revit document to be processed.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("level", "The level that the created Floor Plan refers to.")]
+        [Input("viewName", "Optional, name of the new view.")]
```
vietle-bh, line 104:
```suggestion
        [Input("viewName", "(Optional) Name of the new view.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("scopeBox", "Optional, the Scope Box to attempt to apply to the newly created view.")]
```
vietle-bh, line 105:
```suggestion
        [Input("scopeBox", "(Optional) A Scope Box to attempt to apply to the newly created view.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
```
vietle-bh, line 106:
```suggestion
        [Input("viewTemplateId", "(Optional) A View Template Id to be applied to the view.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("cropRegionVisible", "True if enable crop region visibility.")]
```
vietle-bh, line 107:
```suggestion
        [Input("cropRegionVisible", "True if the crop region should be visible.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("annotationCrop", "True if enable annotation crop visibility.")]
```
vietle-bh, line 108:
```suggestion
        [Input("annotationCrop", "True if the annotation crop should be visible.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Output("viewPlan", "The new view.")]
+        public static View ViewPlan(this Document document, string viewName, Level level, ElementId scopeBoxId = null, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
+        {
+            View result = null;
+
+            ViewFamilyType vft = Query.ViewFamilyType(document, ViewFamily.FloorPlan);
```
vietle-bh, line 114:
```suggestion
            ViewFamilyType viewType = Query.ViewFamilyType(document, ViewFamily.FloorPlan);
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+
+            result = Autodesk.Revit.DB.ViewPlan.Create(document, vft.Id, level.Id);
```
vietle-bh, line 116:
```suggestion
            result = ViewPlan.Create(document, vft.Id, level.Id);
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+            if (viewTemplateId != null)
+            {
+                try
+                {
+                    result.ViewTemplateId = viewTemplateId;
+                }
+                catch (Exception)
+                {
+                    BH.Engine.Base.Compute.RecordWarning("Could not apply the View Template of Id " + viewTemplateId + "'." + ". Please check if it's a valid ElementId.");
+                }
```
vietle-bh, line 127: It can be cleaner to get the view using its ID, check its `IsTemplate` property, and then do the assignment or raise the warning accordingly.

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+            {
+                try
+                {
+#if (REVIT2018 || REVIT2019)
+                    result.ViewName = viewName;
+#else
+                    result.Name = viewName;
+#endif
+                }
+                catch
```
vietle-bh, line 140:
```suggestion
                catch (Autodesk.Revit.Exceptions.ArgumentException)
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+                {
+                    BH.Engine.Base.Compute.RecordWarning("There is already a view named '" + viewName + "'." + " It has been named '" + result.Name + "' instead.");
```
vietle-bh, line 142: Instead of using the new name Revit auto-generates, should we append (1)/(2)/(3)/etc. to the name the user has specified?
If so, a while loop to increment this number until the new name is unique will do. The user can then rename the sheet as required much quicker.

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+            if (scopeBoxId != null)
+            {
+                try
+                {
+                    result.get_Parameter(BuiltInParameter.VIEWER_VOLUME_OF_INTEREST_CROP).Set(scopeBoxId);
+                }
+                catch (Exception)
+                {
+                    BH.Engine.Base.Compute.RecordWarning("Could not apply the Scope Box of Id " + scopeBoxId + "'." + ". Please check if it's a valid ElementId.");
+                }
```
vietle-bh, line 155: It can be cleaner to get the scope box element and check if its category is `OST_VolumeOfInterest`, and then only do the assignment if that's true.

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+                
+            }
+
+            if (cropRegionVisible)
+                result.get_Parameter(BuiltInParameter.VIEWER_CROP_REGION_VISIBLE).Set(1);
+            else
+                result.get_Parameter(BuiltInParameter.VIEWER_CROP_REGION_VISIBLE).Set(0);
```
vietle-bh, line 162: Will `Set(cropRegionVisible? 1 : 0)` work?

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+
+            if (annotationCrop)
+                result.get_Parameter(BuiltInParameter.VIEWER_ANNOTATION_CROP_ACTIVE).Set(1);
+            else
+                result.get_Parameter(BuiltInParameter.VIEWER_ANNOTATION_CROP_ACTIVE).Set(0);
```
vietle-bh, line 167: Will `Set(annotationCrop? 1 : 0)` work?

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -97,6 +97,77 @@ public static View ViewReflectedCeilingPlan(this Document document, Autodesk.Rev
...
+        [Description("Creates and returns a new Reflected Ceiling Plan view in the current Revit file.")]
+        [Input("document", "Revit current document to be processed.")]
+        [Input("level", "The level that the created Floor Plan refers to.")]
+        [Input("viewName", "Optional, name of the new view.")]
+        [Input("scopeBoxId", "Optional, the Scope Box to attempt to apply to the newly created view.")]
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if enable crop region visibility.")]
+        [Input("annotationCrop", "True if enable annotation crop visibility.")]
+        [Output("viewReflectedCeilingPlan", "The new view.")]
+        public static View ViewReflectedCeilingPlan(this Document document, string viewName, Level level, ElementId scopeBoxId = null, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
```
vietle-bh, line 110: Can't we just use `Revit_Core_Engine\Create\View\ViewPlan.cs` and no longer need this method?

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Compute the outline of Title Block drawing area.")]
+        [Input("document", "Revit current document to be processed.")]
```
vietle-bh, line 41:
```suggestion
        [Input("document", "The current Revit document to be processed.")]
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+        [Input("titleBlock", "Title Block symbol to get the outline drawing area from.")]
```
vietle-bh, line 42:
```suggestion
        [Input("titleBlock", "Title Block symbol to get the drawing area outline from.")]
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+        [Output("outline", "Title Block drawing area.")]
```
vietle-bh, line 43:
```suggestion
        [Output("outline", "The Title Block's drawing area.")]
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+        public static Outline DrawingArea(this Document document, FamilySymbol titleBlock)
+        {
+            Document familyDoc = document.EditFamily(titleBlock.Family);
+            List<DetailLine> lines = new FilteredElementCollector(familyDoc).OfCategory(Autodesk.Revit.DB.BuiltInCategory.OST_Lines).WhereElementIsNotElementType().Select(x => x as DetailLine).ToList();
```
vietle-bh, line 47:
```suggestion
            List<DetailLine> lines = new FilteredElementCollector(familyDoc).OfCategory(BuiltInCategory.OST_Lines).WhereElementIsNotElementType().Cast<DetailLine>().ToList();
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            BH.oM.Geometry.CompositeGeometry compositeGeom = new BH.oM.Geometry.CompositeGeometry();
```
vietle-bh, line 49:
```suggestion
            var compositeGeom = new CompositeGeometry();
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            foreach (DetailLine dLine in lines)
+            {
+                BH.oM.Geometry.Line bhomLine = (dLine.Location as LocationCurve).Curve.IFromRevit() as BH.oM.Geometry.Line;
```
vietle-bh, line 53:
```suggestion
                var bhomLine = dLine.GeometryCurve.IFromRevit() as BH.oM.Geometry.Line;
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+                compositeGeom.Elements.Add(bhomLine);
+            }
+
+            BH.oM.Geometry.Point centrePoint = compositeGeom.Bounds().Centre();
+
+            var upLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint.Translate(Vector.YAxis * 10));
+            var rightLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint.Translate(Vector.XAxis * 10));
+            var downLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint.Translate(-Vector.YAxis * 10));
+            var leftLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint.Translate(-Vector.XAxis * 10));
```
vietle-bh, line 62:
```suggestion
            var upLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.YAxis * 10);
            var rightLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.XAxis * 10);
            var downLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.YAxis * 10);
            var leftLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.XAxis * 10);
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            List<BH.oM.Geometry.Point> upPoints = new List<BH.oM.Geometry.Point>();
+            List<BH.oM.Geometry.Point> rightPoints = new List<BH.oM.Geometry.Point>();
+            List<BH.oM.Geometry.Point> downPoints = new List<BH.oM.Geometry.Point>();
+            List<BH.oM.Geometry.Point> leftPoints = new List<BH.oM.Geometry.Point>();
```
vietle-bh, line 67:
```suggestion
            var upPoints = new List<BH.oM.Geometry.Point>();
            var rightPoints = new List<BH.oM.Geometry.Point>();
            var downPoints = new List<BH.oM.Geometry.Point>();
            var leftPoints = new List<BH.oM.Geometry.Point>();
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+                if (leftPoint != null)
+                    leftPoints.Add(leftPoint);
+            }
+
+            var areaUpPoint = upPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaRightPoint = rightPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaDownPoint = downPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaLeftPoint = leftPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+
+            List<BH.oM.Geometry.Point> areaPoints = new List<BH.oM.Geometry.Point> { areaUpPoint, areaRightPoint, areaDownPoint, areaLeftPoint };
```
vietle-bh, line 93:
```suggestion
            var areaPoints = new List<BH.oM.Geometry.Point> { areaUpPoint, areaRightPoint, areaDownPoint, areaLeftPoint };
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,110 @@
...
+            var minY = areaPoints.Select(x => x.Y).OrderBy(x => x).FirstOrDefault();
+            var maxX = areaPoints.Select(x => x.X).OrderByDescending(x => x).FirstOrDefault();
+            var maxY = areaPoints.Select(x => x.Y).OrderByDescending(x => x).FirstOrDefault();
+
+            var minPoint = BH.Engine.Geometry.Create.Point(minX, minY);
+            var maxPoint = BH.Engine.Geometry.Create.Point(maxX, maxY);
+
+            Outline drawingAreaOutline = new Outline(minPoint.ToRevit(), maxPoint.ToRevit());
+
+            return drawingAreaOutline;
```
vietle-bh, line 105: Could be cleaner to do:

```suggestion
            BoundingBox areaBox = areaPoints.Bounds();
            return new Outline(areaBox.Min.ToRevit(), areaBox.Max.ToRevit());
```

Revit_Core_Engine/Compute/PlaceViewOnSheet.cs
michal-pekacki: It would be 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: View Name is not optional - I changed it

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: It doesn't work because there is already ViewPlan.Create method in Revit_Toolkit

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: Much better, thanks

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
michal-pekacki: Good idea, I will add ViewType input and make it as a one method 👍

Revit_Core_Engine/Query/DrawingArea.cs
michal-pekacki: Couldn't remove `Autodesk.Revit.DB` because there is already `BuiltInCategory` method in Revit_Tools

Revit_Core_Engine/Create/Sheet.cs
michal-pekacki: Sheet names can be duplicated in the model, so it is not needed, but it has to be done for sheet numbers that can't be.

Revit_Core_Engine/Create/Sheet.cs
michal-pekacki: Good idea, will do it 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: Good idea, will do it 👍

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,83 @@
...
+                        }
+
+                        result.SheetNumber = sheetNumber;
+                        break;
+                    }
+                    catch
+                    {
+                        number++;
+                    }
+                }
```
vietle-bh, line 72: I often feel nervous when seeing `while (true)` 😅 and also avoid `try catch` whenever I can.

In this case, it can `RecordWarning` multiple times if `number` has to increment a few times to make a unique sheet number. This will give a longer and less clear error log.

What I would do is create a method just for checking if a sheet number already exists (`IsExistingSheetNumber`) and then do something like this:
```c#
int number = 0;
string uniqueNumber = sheetNumber;

while (uniqueNumber.IsExistingSheetNumber())
{
    number++;
    uniqueNumber = $"{sheetNumber} ({number})";
}

result.SheetNumber = uniqueNumber;
if (uniqueNumber != sheetNumber)
{
    BH.Engine.Base.Compute.RecordWarning($"There is already a sheet named '{sheetNumber}'. It has been named '{uniqueNumber}' instead.");
}
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,83 @@
...
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("sheetName", "Name of the new sheet.")]
+        [Input("sheetNumber", "Number of the new sheet.")]
+        [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
+        [Output("viewSheet", "The new sheet.")]
+        public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
```
vietle-bh, line 41: I would consider Sheet just another type of view and move this method to a new file at `Revit_Core_Engine\Create\View\Sheet.cs`

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,107 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+#else
+                        result.Name = viewName;
+#endif
+                        break;
+                    }
+                    catch
+                    {
+                        number++;
+                    }
+                }
```
vietle-bh, line 173: To avoid checking the Revit version before setting the view name, you can do `result.get_Parameter(BuiltInParameter.VIEW_NAME).Set(newName);`

Also, similar to sheet number, it would be nice to avoid the `while(true)` and `try catch` with something like this:
```c#
int number = 0;
string uniqueName = viewName;

while (uniqueName.IsExistingViewName())
{
    number++;
    uniqueName = $"{viewName} ({number})";
}

result.get_Parameter(BuiltInParameter.VIEW_NAME).Set(uniqueName);
if (uniqueName != viewName)
{
    BH.Engine.Base.Compute.RecordWarning($"There is already a view named '{viewName}'. It has been named '{uniqueName}' instead.");
}
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -97,6 +97,107 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("viewName", "Name of the new view.")]
+        [Input("viewFamily", "View Family of the View Type. The default is FloorPlan.")]
+        [Input("scopeBox", "(Optional) A Scope Box to attempt to apply to the newly created view.")]
+        [Input("viewTemplateId", "(Optional) A View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("viewPlan", "The new view.")]
+        public static View ViewPlan(this Document document, Level level, string viewName, ViewFamily viewFamily = ViewFamily.FloorPlan,  ElementId scopeBoxId = null, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
+        {
+            View result = null;
```
vietle-bh, line 113:
```suggestion
            View newView = null;
```

Revit_Core_Engine/Create/Sheet.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            ViewSheet result = ViewSheet.Create(document, titleBlockId);
```
vietle-bh, line 43:
```suggestion
            ViewSheet newSheet = ViewSheet.Create(document, titleBlockId);
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,18 +39,18 @@ public static partial class Create
         [Input("viewName", "Optional, name of the new view.")]
         [Input("cropBox", "Optional, the crop region to attempt to apply to the newly created view.")]
         [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
-        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]        
-        [Output("viewReflectedCeilingPlan", "The new view.")]        
+        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]
+        [Output("viewReflectedCeilingPlan", "The new view.")]
         public static View ViewReflectedCeilingPlan(this Document document, Autodesk.Revit.DB.Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
```
vietle-bh, line 44:
```suggestion
        public static View ViewReflectedCeilingPlan(this Document document,Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,18 +39,18 @@ public static partial class Create
...
         {
             View result = null;
```
vietle-bh, line 46:
```suggestion
            View newView = null;
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,18 +39,18 @@ public static partial class Create
...
 
             ViewFamilyType vft = Query.ViewFamilyType(document, ViewFamily.CeilingPlan);
```
vietle-bh, line 48:
```suggestion
            ViewFamilyType viewFamilyType = Query.ViewFamilyType(document, ViewFamily.CeilingPlan);
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,18 +39,18 @@ public static partial class Create
...
 
             result = Autodesk.Revit.DB.ViewPlan.Create(document, vft.Id, level.Id);
-            
+
             Modify.SetViewDetailLevel(result, viewDetailLevel);
-            
+
             if (cropBox != null)
             {
                 try
```
vietle-bh, line 56: You can check the view's `ViewTemplateId` property and only apply the crop box if the template is in use. Then, the `RecordWarning` in this `try catch` can only mention "...the crop box is a valid geometry..." to make it more specific for the user.

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -76,7 +76,7 @@ public static View ViewReflectedCeilingPlan(this Document document, Autodesk.Rev
                     BH.Engine.Base.Compute.RecordWarning("Could not apply the View Template of Id " + viewTemplateId + "'." + ". Please check if it's a valid ElementId.");
```
vietle-bh, line 76: Before setting this property, we should obtain the element at the view template ID and check if it's an actual view template.

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -92,7 +92,7 @@ public static View ViewReflectedCeilingPlan(this Document document, Autodesk.Rev
                     BH.Engine.Base.Compute.RecordWarning("There is already a view named '" + viewName + "'." + " It has been named '" + result.Name + "' instead.");
```
vietle-bh, line 92: This can be avoided by using the same approach suggested for sheet number and plan view name. The same check should take place here before setting the new view name.

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+        [Description("Compute the outline of Title Block drawing area.")]
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("titleBlock", "Title Block symbol to get the drawing area outline from.")]
+        [Output("outline", "The Title Block's drawing area.")]
+        public static Outline DrawingArea(this Document document, FamilySymbol titleBlock)
+        {
+            Document familyDoc = document.EditFamily(titleBlock.Family);
+            List<DetailLine> lines = new FilteredElementCollector(familyDoc).OfCategory(Autodesk.Revit.DB.BuiltInCategory.OST_Lines).WhereElementIsNotElementType().Cast<DetailLine>().ToList();
+
+            var compositeGeom = new BH.oM.Geometry.CompositeGeometry();
```
vietle-bh, line 49:
```suggestion
            var compositeGeom = new CompositeGeometry();
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            foreach (DetailLine dLine in lines)
+            {
+                var bhomLine = (dLine.Location as LocationCurve).Curve.IFromRevit() as BH.oM.Geometry.Line;
```
vietle-bh, line 53:
```suggestion
                var bhomLine = dLine.GeometryCurve.IFromRevit() as BH.oM.Geometry.Line;
```

vietle-bh (commented):
In addition to the comments below, we should switch all `BH.Engine.Base.Compute.RecordWarning` call to using the `$` string interpolation syntax instead of `string + string + string etc.` :bowtie:

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
michal-pekacki: As discussed, view templates and cropbox settings are not connected, so I will change the warning message.

Revit_Core_Engine/Create/View/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("sheetName", "Name of the new sheet.")]
+        [Input("sheetNumber", "Number of the new sheet.")]
+        [Input("viewTemplateId", "The Title Block Id to be applied to the sheet.")]
+        [Output("newSheet", "The new sheet.")]
+        public static ViewSheet Sheet(this Document document, string sheetName, string sheetNumber, ElementId titleBlockId)
+        {
+            List<string> sheetNumbersInModel = new FilteredElementCollector(document).OfClass(typeof(ViewSheet)).WhereElementIsNotElementType().Select(x => (x as ViewSheet).SheetNumber).ToList();
```
pawelbaran, line 45: `WhereElementIsNotElementType` is not needed I think - `ViewSheet` will never be a type. Also casting could be slightly more elegant I think

```suggestion
            List<string> sheetNumbersInModel = new FilteredElementCollector(document).OfClass(typeof(ViewSheet)).Cast<ViewSheet>().Select(x => x.SheetNumber).ToList();
```

Revit_Core_Engine/Create/View/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
```
pawelbaran, line 37: Are you sure this description is applicable to this method? 🙈

Revit_Core_Engine/Create/View/Sheet.cs
```diff
@@ -0,0 +1,79 @@
...
+                while (sheetNumbersInModel.Contains(uniqueNumber))
+                {
+                    number++;
+                    uniqueNumber = $"{sheetNumber} ({number})";
+                }
+
+                newSheet.SheetNumber = uniqueNumber;
+                if (uniqueNumber != sheetNumber)
+                {
+                    BH.Engine.Base.Compute.RecordWarning($"There is already a sheet named '{sheetNumber}'. It has been named '{uniqueNumber}' instead.");
```
pawelbaran, line 67:
```suggestion
                    BH.Engine.Base.Compute.RecordWarning($"Sheet named '{sheetNumber}' already exists in the document. Newly created sheet has been named '{uniqueNumber}' instead.");
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -79,24 +88,135 @@ public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level leve
...
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("level", "The level that the created Floor Plan refers to.")]
+        [Input("viewName", "Name of the new view.")]
+        [Input("viewFamily", "View Family of the View Type. The default is FloorPlan.")]
+        [Input("scopeBox", "(Optional) A Scope Box to attempt to apply to the newly created view.")]
+        [Input("viewTemplateId", "(Optional) A View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("newView", "The new view.")]
+        public static View ViewPlan(this Document document, Level level, string viewName, ViewFamily viewFamily = ViewFamily.FloorPlan,  ElementId scopeBoxId = null, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
```
pawelbaran, line 126: Now when I look at this method and the other one that creates view plan - shouldn't both have same inputs, the only difference being scope box vs crop box? So both would have `viewDetailLevel`, `viewFamily`, `cropRegionVisible`, `annotationCrop` etc. Then it would become one single method more or less, only outline being a different query for each. Or can't it be simplified to this extent?

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,66 +39,73 @@ public static partial class Create
...
                 {
                     BH.Engine.Base.Compute.RecordWarning("Could not apply the View Template of Id " + viewTemplateId + "'." + ". Please check if it's a valid ElementId.");
                 }
             }
-            
+
             if (!string.IsNullOrEmpty(viewName))
             {
-                try
+                int number = 0;
```
pawelbaran, line 88: Shouldn't this entire section become a dedicated method, something around `SetViewName`? Assuming that the process looks the same for each view type.

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            foreach (DetailLine dLine in lines)
+            {
+                var bhomLine = dLine.GeometryCurve.IFromRevit() as BH.oM.Geometry.Line;
+                compositeGeom.Elements.Add(bhomLine);
+            }
+
+            BH.oM.Geometry.Point centrePoint = compositeGeom.Bounds().Centre();
+
+            var upLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.YAxis * 10);
```
pawelbaran, line 59: Why do you multiply by factor 10? Is that because the sheet will never be larger than 10x10m?

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                var leftPoint = leftLine.LineIntersection(bhomLine);
+                if (leftPoint != null)
+                    leftPoints.Add(leftPoint);
+            }
+
+            var areaUpPoint = upPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaRightPoint = rightPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaDownPoint = downPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
+            var areaLeftPoint = leftPoints.OrderBy(x => x.Distance(centrePoint)).FirstOrDefault();
```
pawelbaran, line 91: Wouldn't that work?

```suggestion
            var areaUpPoint = upPoints.OrderBy(x => x.Y).FirstOrDefault();
            var areaRightPoint = rightPoints.OrderBy(x => x.X).FirstOrDefault();
            var areaDownPoint = downPoints.OrderByDescending(x => x.Y).FirstOrDefault();
            var areaLeftPoint = leftPoints.OrderByDescending(x => x.X).FirstOrDefault();
```

Revit_Core_Engine/Query/DrawingArea.cs
michal-pekacki: Yes 👍

I changed it to `double.MaxValue`

michal-pekacki: > Yes 👍
> 
> I changed it to `double.MaxValue`

Moved it back to `10` because `double.MaxValue` doesn't work.

Revit_Core_Engine/Create/View/ViewPlan.cs
pawelbaran: I will let myself play the devil's advocate one more time 😈 

1. You would have 2 methods, one taking `CurveLoop` and the other `Element`, all other inputs being the same - but then there would be a query extracting the outline of loop and scope box, to be passed to a common method taking the extracted outline and other inputs. Would that work?
2. You could always leave it empty for your purpose, am I correct?
3. I think there is not too many calls to it, should be easy enough to update the code and add versioning

michal-pekacki: I still think it is not good idea to merge it like this because `Scope Box` and `Crop Box` are totally different:
- to set `Scope Box` you need to change `BuiltInParameter.VIEWER_VOLUME_OF_INTEREST_CROP` parameter
- to set `Crop Box` you need to use `SetCropShape()` from `ViewCropRegionShapeManager`
- If you use `Scope Box`, you can't change `Crop Box`

I would stay with two separate methods that will be used in different cases.

vietle-bh: How about having the new method call the existing method with `cropBox = null`, then apply the scope box and anything extra? 

To make that work, we'll probably need to add the `ViewFamily viewFamily = ViewFamily.FloorPlan` input to the existing method though, and do versioning... 😅 

image

vietle-bh: Unless you only do the `if (!(viewFamily == ViewFamily.FloorPlan || viewFamily == ViewFamily.CeilingPlan || viewFamily == ViewFamily.AreaPlan || viewFamily == ViewFamily.StructuralPlan))` check in the new method. In that case no need to do versioning?

Revit_Core_Engine/Query/DrawingArea.cs
pawelbaran: Yeah, that's what I thought when saw your reply about max value 😉 10 is perfectly fine, I do not think we will ever produce sheets of size of a sail

Revit_Core_Engine/Create/View/ViewPlan.cs
michal-pekacki: I combined these methods and added `viewFamily` input to existing. In the new method I call the existing one with `cropBox = null`.

pawelbaran (changes requested):

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -40,63 +43,135 @@ public static partial class Create
...
+        [PreviousVersion("6.1", "BH.Revit.Engine.Core.Create.ViewPlan(Autodesk.Revit.DB.Document, Autodesk.Revit.DB.Level, System.String, Autodesk.Revit.DB.ViewFamily, Autodesk.Revit.DB.CurveLoop, Autodesk.Revit.DB.ElementId, Autodesk.Revit.DB.ViewDetailLevel)")]
+        public static View ViewPlan(this Document document, Level level, string viewName = null, ViewFamily viewFamily = ViewFamily.FloorPlan, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
         {
-            View result = null;
+            View newView = null;
 
-            ViewFamilyType vft = Query.ViewFamilyType(document, ViewFamily.FloorPlan);
+            if (!(viewFamily == ViewFamily.FloorPlan || viewFamily == ViewFamily.CeilingPlan || viewFamily == ViewFamily.AreaPlan || viewFamily == ViewFamily.StructuralPlan))
+            {
+                BH.Engine.Base.Compute.RecordWarning($"Could not create View of type '{viewFamily}'. It has to be a FloorPlan, CeilingPlan, AreaPlan, or StructuralPlan ViewType.");
```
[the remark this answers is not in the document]
michal-pekacki: Changed 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -40,63 +43,135 @@ public static partial class Create
         [Input("cropBox", "Optional, the crop region to attempt to apply to the newly created view.")]
         [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
         [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]        
-        [Output("viewPlan", "The new view.")]        
-        public static View ViewPlan(this Document document, Autodesk.Revit.DB.Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
+        [Output("newView", "The new view.")]
+        [PreviousVersion("6.1", "BH.Revit.Engine.Core.Create.ViewPlan(Autodesk.Revit.DB.Document, Autodesk.Revit.DB.Level, System.String, Autodesk.Revit.DB.ViewFamily, Autodesk.Revit.DB.CurveLoop, Autodesk.Revit.DB.ElementId, Autodesk.Revit.DB.ViewDetailLevel)")]
```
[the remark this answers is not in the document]
michal-pekacki: Moved to the top 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -40,63 +43,135 @@ public static partial class Create
...
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("level", "The level that the created Floor Plan refers to.")]
+        [Input("viewName", "Name of the new view.")]
+        [Input("viewFamily", "View Family of the View Type. The default is FloorPlan.")]
+        [Input("scopeBox", "(Optional) A Scope Box to attempt to apply to the newly created view.")]
+        [Input("viewTemplateId", "(Optional) A View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("newView", "The new view.")]
+        public static View ViewPlan(this Document document, Level level, string viewName, ViewFamily viewFamily = ViewFamily.FloorPlan,  ElementId scopeBoxId = null, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
```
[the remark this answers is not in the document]
michal-pekacki: `cropBox` and `scopeBox` inputs are optional and they can't be next to `document`. I can place them next to `viewName` but there will be no big difference.

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -40,63 +43,135 @@ public static partial class Create
...
                 {
-                    BH.Engine.Base.Compute.RecordWarning("There is already a view named '" + viewName + "'." + " It has been named '" + result.Name + "' instead.");
+                    BH.Engine.Base.Compute.RecordWarning($"Could not apply the Scope Box of Id '{scopeBoxId}'. Please check if it's a valid ElementId.");
                 }
             }
-            
-            return result;
+
+            newView.get_Parameter(BuiltInParameter.VIEWER_CROP_REGION_VISIBLE).Set(cropRegionVisible? 1: 0);
+            newView.get_Parameter(BuiltInParameter.VIEWER_ANNOTATION_CROP_ACTIVE).Set(annotationCrop? 1: 0);
```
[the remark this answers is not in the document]
michal-pekacki: I don't know if it's possible, but this is Revit and it's always worth to add such a check 👍

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -39,66 +39,60 @@ public static partial class Create
         [Input("viewName", "Optional, name of the new view.")]
         [Input("cropBox", "Optional, the crop region to attempt to apply to the newly created view.")]
         [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
-        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]        
-        [Output("viewReflectedCeilingPlan", "The new view.")]        
-        public static View ViewReflectedCeilingPlan(this Document document, Autodesk.Revit.DB.Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
+        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]
+        [Output("viewReflectedCeilingPlan", "The new view.")]
+        public static View ViewReflectedCeilingPlan(this Document document, Level level, string viewName = null, CurveLoop cropBox = null, ElementId viewTemplateId = null, ViewDetailLevel viewDetailLevel = ViewDetailLevel.Coarse)
```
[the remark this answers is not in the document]
michal-pekacki: `[ToBeRemoved("6.1")]` attribute added 👍

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+        [Description("Compute the outline of Title Block drawing area.")]
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("titleBlock", "Title Block symbol to get the drawing area outline from.")]
+        [Output("outline", "The Title Block's drawing area.")]
+        public static Outline DrawingArea(this Document document, FamilySymbol titleBlock)
+        {
+            Document familyDoc = document.EditFamily(titleBlock.Family);
+            List<DetailLine> lines = new FilteredElementCollector(familyDoc).OfCategory(Autodesk.Revit.DB.BuiltInCategory.OST_Lines).WhereElementIsNotElementType().Where(x => x is DetailLine).Cast<DetailLine>().ToList();
+
+            var compositeGeom = new CompositeGeometry();
```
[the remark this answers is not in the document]
michal-pekacki: I used `CompositeGeometry` to easily get the center point in the next lines.

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,101 @@
...
+        /***************************************************/
+
+        [Description("Compute the outline of Title Block drawing area.")]
+        [Input("document", "The current Revit document to be processed.")]
+        [Input("titleBlock", "Title Block symbol to get the drawing area outline from.")]
+        [Output("outline", "The Title Block's drawing area.")]
+        public static Outline DrawingArea(this Document document, FamilySymbol titleBlock)
+        {
+            Document familyDoc = document.EditFamily(titleBlock.Family);
+            List<DetailLine> lines = new FilteredElementCollector(familyDoc).OfCategory(Autodesk.Revit.DB.BuiltInCategory.OST_Lines).WhereElementIsNotElementType().Where(x => x is DetailLine).Cast<DetailLine>().ToList();
```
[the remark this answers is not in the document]
michal-pekacki: Can't use `DetailLine` in this way:

image

Revit_Core_Engine/Create/View/ViewPlan.cs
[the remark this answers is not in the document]
FraserGreenroyd: There isn't any agreed convention but most people (except @pawelbaran it seems) have gone with putting it close to the method signature for easier human verification that it's matching.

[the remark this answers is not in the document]
pawelbaran: Apparently I like making life harder... 🙈

Revit_Core_Engine/Create/View/ViewPlan.cs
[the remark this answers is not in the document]
pawelbaran: Good point, in this case it could be worth putting them as 1st optional, but no strong opinion here - I will mark it as resolved 👍

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -33,70 +36,149 @@ public static partial class Create
         /****               Public Methods              ****/
         /***************************************************/
 
+        [PreviousVersion("6.1", "BH.Revit.Engine.Core.Create.ViewPlan(Autodesk.Revit.DB.Document, Autodesk.Revit.DB.Level, System.String, Autodesk.Revit.DB.ViewFamily, Autodesk.Revit.DB.CurveLoop, Autodesk.Revit.DB.ElementId, Autodesk.Revit.DB.ViewDetailLevel)")]
         [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
```
pawelbaran, line 40:
```suggestion
        [Description("Creates and returns a new Plan view in the current Revit file.")]
```

Revit_Core_Engine/Create/View/ViewPlan.cs
```diff
@@ -33,70 +36,149 @@ public static partial class Create
...
+
+            if (!newView.get_Parameter(BuiltInParameter.VIEWER_ANNOTATION_CROP_ACTIVE).Set(annotationCrop ? 1 : 0))
+                BH.Engine.Base.Compute.RecordWarning($"Could not set the annotation crop in the view. Parameter is ready-only.");
+
+            return newView;
+        }
+
+        /***************************************************/
+
+        [Description("Creates and returns a new Floor Plan view in the current Revit file.")]
```
pawelbaran, line 114:
```suggestion
        [Description("Creates and returns a new Plan view in the current Revit file.")]
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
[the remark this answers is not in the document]
pawelbaran: Could you please add a description that this method is to be replaced with `Create.ViewPlan`? For others to know how to replace the method and us to know what to do in 6 months from now 😉

Revit_Core_Engine/Create/View/Viewport.cs
```diff
@@ -0,0 +1,70 @@
...
+        /***************************************************/
+
+        [Description("Places view on sheet at specified position and rotation using the specified viewport type.")]
+        [Input("sheet", "Sheet to place the view on.")]
+        [Input("view", "View to be placed on sheet.")]
+        [Input("viewportTypeId", "Id of the viewport type to be used.")]
+        [Input("viewportPlacementPoint", "Placement point of the viewport on the sheet.")]
+        [Input("viewportRotation", "Rotation type of the viewport. The default is None.")]
+        [Output("viewPort", "The new viewport.")]
+        public static Viewport Viewport (this ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = ViewportRotation.None)
```
pawelbaran, line 43:
```suggestion
        public static Viewport Viewport(this ViewSheet sheet, View view, ElementId viewportTypeId, XYZ viewportPlacementPoint = null, ViewportRotation viewportRotation = ViewportRotation.None)
```

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            BH.oM.Geometry.Point centrePoint = compositeGeom.Bounds().Centre();
+            
+            var leftLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.XAxis * 10);
+            var rightLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.XAxis * 10);
+            var upLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.YAxis * 10);
+            var downLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.YAxis * 10);
+
+            double minX = double.MinValue;
```
pawelbaran, line 66: 2nd thought on that - I would consider using other variable name, e.g. `leftBound` etc. Because in fact we are not looking for minimum X, rather max X to the left from centre point. Sorry for the mess

pawelbaran (commented):
Last few remarks left 👍

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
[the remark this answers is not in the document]
michal-pekacki: Added 👍

Revit_Core_Engine/Query/DrawingArea.cs
```diff
@@ -0,0 +1,98 @@
...
+            
+            var leftLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.XAxis * 10);
+            var rightLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.XAxis * 10);
+            var upLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint + Vector.YAxis * 10);
+            var downLine = BH.Engine.Geometry.Create.Line(centrePoint, centrePoint - Vector.YAxis * 10);
+
+            double leftBound = double.MinValue;
+            double rightBound = double.MaxValue;
+            double downBound = double.MinValue;
+            double upBound = double.MaxValue;
```
pawelbaran, line 69:
```suggestion
            double bottomBound = double.MinValue;
            double topBound = double.MaxValue;
```

Revit_Core_Engine/Create/View/ViewReflectedCeilingPlan.cs
```diff
@@ -33,72 +33,67 @@ public static partial class Create
         /****               Public Methods              ****/
         /***************************************************/
 
+        [ToBeRemoved("6.1","This method is to be replaced with Create.ViewPlan.")]
```
pawelbaran, line 36:
```suggestion
        [ToBeRemoved("6.1","Calls to this method are to be replaced with Create.ViewPlan with 'viewFamily' parameter equal to ViewFamily.CeilingPlan.")]
```

pawelbaran (changes requested):
Last of the last from my side 👍

pawelbaran (dismissed):
Happy to approve from code side 👍 To be merged after functional verification.

pawelbaran (approved):
Still happy with the code 👍

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance

pawelbaran (approved):
Still happy with the changes after the last commit

pawelbaran:
@BHoMBot check serialisation
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check ready-to-merge

PR #1322, opened 2023-03-15
Performance of basic vector converts fixed

### LINKED ISSUE #1303: Performance of converts could be speeded up significantly by caching unit type
#### Description:

Currently on every unit convert `BHoMUnitType` method is executed, resulting in redundant 2 `foreach` loops and one Linq query that seem to take up majority of the entire execution time. This becomes an issue in case of performance-sensitive converts. Quick solution is available - it should be relatively simple to cache units for each unit type instead of querying them every time.

Another level of optimisation would be to store the actual conversion factors as doubles, but this would make the code less readable and the change more painful - not sure if it is worth, depends on how much will the performance improve.

pawelbaran:

Partial solution to #1303

### DIFF
--- a/Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
+++ b/Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
@@ -53,7 +53,7 @@ namespace BH.Revit.Engine.Core
             if (xyz == null)
                 return null;
 
-            return new oM.Geometry.Point { X = xyz.X * m_LengthToSi, Y = xyz.Y * m_LengthToSi, Z = xyz.Z * m_LengthToSi };
+            return new oM.Geometry.Point { X = xyz.X * m_LengthToSI, Y = xyz.Y * m_LengthToSI, Z = xyz.Z * m_LengthToSI };
         }
 
 
@@ -62,7 +62,7 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         // Optimisation to avoid calling ToSI(SpecTypeId.Length) every time as it causes massive overhead
-        private static double m_LengthToSi = 1.0.ToSI(SpecTypeId.Length);
+        private static double m_LengthToSI = 1.0.ToSI(SpecTypeId.Length);
 
         /***************************************************/
     }
--- a/Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
+++ b/Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
@@ -53,8 +53,16 @@ namespace BH.Revit.Engine.Core
             if (xyz == null)
                 return null;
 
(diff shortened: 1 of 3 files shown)


pawelbaran:
@BHoMBot check serialisation
@BHoMBot check null-handling

Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
```diff
@@ -53,9 +53,17 @@ public static oM.Geometry.Point PointFromRevit(this XYZ xyz)
...
+            return new oM.Geometry.Point { X = xyz.X * m_LengthToSi, Y = xyz.Y * m_LengthToSi, Z = xyz.Z * m_LengthToSi };
         }
 
+
+        /***************************************************/
+        /****              Private fields               ****/
+        /***************************************************/
+
+        // Optimisation to avoid calling ToSI(SpecTypeId.Length) every time as it causes massive overhead
+        private static double m_LengthToSi = 1.0.ToSI(SpecTypeId.Length);
```
vietle-bh, line 65: Would look more consistent...
```suggestion
        private static double m_LengthToSI = 1.0.ToSI(SpecTypeId.Length);
```

Revit_Core_Engine/Convert/Geometry/ToRevit/XYZ.cs
```diff
@@ -47,9 +47,17 @@ public static XYZ ToRevit(this oM.Geometry.Point point)
...
+            return new XYZ(vector.X * m_LengthFromSi, vector.Y * m_LengthFromSi, vector.Z * m_LengthFromSi);
         }
 
+
+        /***************************************************/
+        /****              Private fields               ****/
+        /***************************************************/
+
+        // Optimisation to avoid calling FromSI(SpecTypeId.Length) every time as it causes massive overhead
+        private static double m_LengthFromSi = 1.0.FromSI(SpecTypeId.Length);
```
vietle-bh, line 59: Would look more consistent...
```suggestion
        private static double m_LengthFromSI = 1.0.FromSI(SpecTypeId.Length);
```

vietle-bh (commented):
Huge performance improvement ❤️ 

image

Revit_Core_Engine/Convert/Geometry/FromRevit/Point.cs
pawelbaran: Not sure if using multiple capitals in a row is the right convention, but we did it so far, so applied the change to keep convention.

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance

pawelbaran:
@BHoMBot check serialisation
@BHoMBot check null-handling
@BHoMBot check versioning
@BHoMBot check installer

vietle-bh (approved):
Works well with great naming 👍

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #1357, opened 2023-05-22
Add `Create.Elevation`, `Query.Points` methods, update `Create.View3D`

### LINKED ISSUE #1354: Add Elevation create method
#### Description:

As in the title.

michal-pekacki:

Closes #1354 

- added `Create.Elevation` method to create elevation views
- added `Query.Points` method to get all bounding box points
- updated `Create.View3D` method with CropBox alignment


Revit_Core_Engine/Create/View/Viewport.cs
```diff
@@ -44,6 +44,9 @@ public static Viewport Viewport(this ViewSheet sheet, View view, ElementId viewp
         {
             Document document = sheet.Document;
 
+            if (viewportPlacementPoint == null)
```
pawelbaran, line 47: I would add `RecordNote` here. Also could you please test how does this change affect the behaviour of pushing the viewport via Revit adapter? Something to be added as a test script for this PR

Revit_Core_Engine/Create/View/Viewport.cs
```diff
@@ -55,10 +58,11 @@ public static Viewport Viewport(this ViewSheet sheet, View view, ElementId viewp
             viewPort.ChangeTypeId(viewportTypeId);
             viewPort.Rotation = viewportRotation;
 
-            document.Regenerate();
-
             if (viewportPlacementPoint != null)
+            {
```
pawelbaran, line 62: `viewportPlacementPoint` will never be null now

Revit_Core_Engine/Query/Points.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Queries edge points of BoundingBoxXYZ.")]
```
pawelbaran, line 36:
```suggestion
        [Description("Queries corner points of BoundingBoxXYZ.")]
```

Revit_Core_Engine/Query/Points.cs
```diff
@@ -0,0 +1,61 @@
...
+        [Input("bbox", "BoundingBoxXYZ to get the points from.")]
+        [Output("edgeCurves", "Edge points of the BoundingBoxXYZ.")]
+        public static List<XYZ> Points(this BoundingBoxXYZ bbox)
+        {
```
pawelbaran, line 40: I believe you should be transforming the points by bbox.Transform if it is not `null`

pawelbaran (changes requested):

Revit_Core_Engine/Create/View/ViewElevations.cs
```diff
@@ -0,0 +1,108 @@
...
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
+        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("newView", "The new view.")]
+        public static List<ViewSection> ViewElevations(this Document document, string elevationName, XYZ elevationMarkerLocation, View referenceViewPlan, int numberOfElevations = 4, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
+        {
+            if (numberOfElevations < 1 || numberOfElevations > 4)
+            {
+                BH.Engine.Base.Compute.RecordWarning($"Number of elevation views should be between 1 and 4.");
```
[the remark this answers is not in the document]
pawelbaran: Ahh ok, got it - I would then consider adding an overload to generate a marker plus single view, i.e. `ViewElevation`, to unlock the cases where the walls are not orthogonal. Or would that not make sense?

Revit_Core_Engine/Create/View/Viewport.cs
michal-pekacki: Right, I removed this check 👍

Revit_Core_Engine/Create/View/Viewport.cs
michal-pekacki: I removed this check, if `viewportPlacementPoint` is null, `Viewport` will be placed in default position `(0, 0, 0)`. I added `RecordNote`.

Revit_Core_Engine/Create/View/ViewElevations.cs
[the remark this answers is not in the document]
michal-pekacki: It is not possible to create elevation view without the elevation marker. For not orthogonal walls created elevation views should be rotated but there still must be maximum 4 of them.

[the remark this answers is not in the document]
pawelbaran: Yeah, so as discussed offline, I would consider adding a method that would create a marker + single view, i.e. for a 5-walled room you would have 5 markers and 5 elevations, one per each. Not sure if that makes sense from practical point of view, so please dismiss if it does not make sense 😃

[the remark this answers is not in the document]
vietle-bh: I think it's necessary to create 1 marker per wall because many rooms have more than 4 walls.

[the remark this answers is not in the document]
michal-pekacki: Ok, I think have a solution for this - I will rename this method to `Create.ElevationMarker` and it will always create one marker with up to 4 elevation views. For more than 4 walls or other non-rectangular rooms we should use this method multiple times - it depends on the case we have. What do you think @pawelbaran @vietle-bh?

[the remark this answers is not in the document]
vietle-bh: How about we first check if the room boundaries form a rectangle? If yes, we create 1 marker for all 4 elevations. If not, we create 1 marker per wall, each has only 1 elevation on it :bowtie:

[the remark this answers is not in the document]
FraserGreenroyd: > How about we first check if the room boundaries form a rectangle? If yes, we create 1 marker for all 4 elevations. If not, we create 1 marker per wall, each has only 1 elevation on it :bowtie:

I think that check will add unnecessary overhead particularly when calling the method multiple times.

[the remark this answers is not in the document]
michal-pekacki: I think there should be no `Room` or `Space` outline input in this method.

michal-pekacki:
@BHoMBot check compliance

Revit_Core_Engine/Create/View/Elevation.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        [Description("Creates Elevation view based on given line. Elevation marker will be placed in the middle of the line.")]
+        [Input("document", "Revit current document to be processed.")]
+        [Input("elevationLine", "Location line of the elevation view and the bottom line for the crop region shape.")]
+        [Input("referenceViewPlan", "ViewPlan that elevation marker is referenced to.")]
+        [Input("elevationName", "Name of the elevation view.")]
+        [Input("depth", "Depth of the elevation view.")]
+        [Input("height", "Height of the elevation view.")]
+        [Input("offset", "Offset that will be added to elevation depth and CropBox shape.")]
+        [Input("elevationMarek", "Elevation marker that has been created (out parameter).")]
```
pawelbaran, line 47: Marek 😶‍🌫️

Revit_Core_Engine/Create/View/Elevation.cs
```diff
@@ -0,0 +1,187 @@
...
+        /***************************************************/
+
+        [Description("Creates Elevation view based on given line. Elevation marker will be placed in the middle of the line.")]
+        [Input("document", "Revit current document to be processed.")]
+        [Input("elevationLine", "Location line of the elevation view and the bottom line for the crop region shape.")]
+        [Input("referenceViewPlan", "ViewPlan that elevation marker is referenced to.")]
+        [Input("elevationName", "Name of the elevation view.")]
+        [Input("depth", "Depth of the elevation view.")]
+        [Input("height", "Height of the elevation view.")]
+        [Input("offset", "Offset that will be added to elevation depth and CropBox shape.")]
```
pawelbaran, line 46: Wondering if it would not be worth splitting offset into 3 values: height, width and depth?

Revit_Core_Engine/Create/View/Elevation.cs
```diff
@@ -0,0 +1,187 @@
...
+            ElementTransformUtils.RotateElement(doc, elevationMarker.Id, axis, angle);
+
+            elevationView.ModifyElevation(elevationName, elevationLine, depth, height, offset, viewTemplateId, cropRegionVisible, annotationCrop);
+
+            return elevationView;
+        }
+
+        /***************************************************/
+
+        [Description("Creates elevation views referenced do elevation marker in the current Revit file.")]
```
pawelbaran, line 74:
```suggestion
        [Description("Creates elevation view referenced do elevation marker in the current Revit file.")]
```

Revit_Core_Engine/Create/View/Elevation.cs
```diff
@@ -0,0 +1,187 @@
...
+        [Input("elevationMarker", "Elevation marker to which elevation views will be referenced.")]
```
pawelbaran, line 75:
```suggestion
        [Input("elevationMarker", "Elevation marker to which elevation view will be referenced.")]
```

Revit_Core_Engine/Create/View/Elevation.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            return newElevation;
+        }
+
+        /***************************************************/
+        /****              Private Methods              ****/
+        /***************************************************/
+
+        [Description("Creates elevation marker in the current Revit file.")]
+        private static ElevationMarker ElevationMarker(this XYZ elevationMarkerLocation, View referenceViewPlan)
```
pawelbaran, line 102: Wouldn't it be worth for this method to be public?

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -56,6 +57,7 @@ public static View View3D(this Document document, string viewName = null, Boundi
                 try
                 {
                     (result as View3D).SetSectionBox(boundingBoxXyz);
+                    AlignCropBoxToBoundingBox(result, boundingBoxXyz);
```
pawelbaran, line 60: Is it always needed? Or only when there is a cropbox? Will it not crash the existing code?

Revit_Core_Engine/Create/View/Elevation.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Create/View/View3D.cs
michal-pekacki: It is always needed and cropbox has to be active to get the correct result. Without alignment 3DView looks like this:

image

Revit_Core_Engine/Create/View/Elevation.cs
michal-pekacki: I changed `offset` to work only for height and width. Depth is defined as separate input.

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check code-compliance

michal-pekacki:
@BHoMBot check code-compliance

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check compliance

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("elevationView", "Elevation view that has been created.")]
+        public static ViewSection ElevationByLine(this Line elevationLine, string elevationName, View referenceViewPlan, double depth, double height, double offset, out ElevationMarker elevationMarker, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
+        {
+            Document doc = referenceViewPlan.Document;
+
+            var elevationMarkerLocation = elevationLine.Evaluate(0.5, true);
+            var lineDirection = elevationLine.Direction.Normalize();
```
pawelbaran, line 56: You need to check if the line is horizontal (dot product with global Z < tolerance), if not then warning (note?) plus making the line horizontal - without that you may get wrong `angle` value

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
```diff
@@ -0,0 +1,80 @@
...
+            var markerDirection = lineDirection.VectorFromRevit().Rotate(-Math.PI / 2, BH.oM.Geometry.Vector.ZAxis).ToRevit();
+            var angle = -markerDirection.AngleTo(XYZ.BasisX);
+
+            if (Math.Abs(angle) < Tolerance.Angle)
```
pawelbaran, line 60: I would check against other rotations as well - see suggestion below (but written off the top of my head, testing needed plus probably can be done smarter). Alternatively you can check based on dot product with global X and Y, the code should be simpler then?
```suggestion
            double absAngle = Math.Abs(angle);
            double refAngle = 0;
            while (refAngle - Tolerance.Angle <= absAngle)
            {
                if (Math.Abs(absAngle - refAngle) < Tolerance.Angle)
                    {
                        if (angle > 0)
                            angle = refAngle;
                        else
                            angle = -refAngle;
                        
                        break;
                    }
                
                refAngle += Math.PI;
            }
```

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
```diff
@@ -0,0 +1,80 @@
...
+                angle = 0;
+
+            var axis = Line.CreateBound(elevationMarkerLocation, elevationMarkerLocation + XYZ.BasisZ);
```
pawelbaran, line 63: Not overly relevant, but you can use an unbound line as an axis
```suggestion
            var axis = Line.CreateUnound(elevationMarkerLocation, XYZ.BasisZ);
```

Revit_Core_Engine/Modify/Elevation.cs
```diff
@@ -0,0 +1,124 @@
...
+        [Input("elevationName", "Name of the elevation view.")]
+        [Input("bottomLine", "Location line of the elevation view and the bottom line for the crop region shape.")]
+        [Input("depth", "Depth of the elevation view.")]
+        [Input("height", "Height of the elevation view.")]
+        [Input("offset", "Offset that will be added to the elevation view CropBox shape.")]
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("elevationView", "Elevation view that has been modified.")]
+        public static void ModifyElevation(this ViewSection elevation, string elevationName, Line bottomLine, double depth, double height, double offset, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
```
pawelbaran, line 51: I am not a big fan of the method name here - couldn't it be `SetElevationProperties` or similar?

Revit_Core_Engine/Modify/Elevation.cs
```diff
@@ -0,0 +1,124 @@
...
+            //set crop region visibility
+            if (!elevation.get_Parameter(BuiltInParameter.VIEWER_CROP_REGION_VISIBLE).Set(cropRegionVisible ? 1 : 0))
+                BH.Engine.Base.Compute.RecordWarning($"Could not set the crop region visibility in the view. Parameter is ready-only.");
+
+            //set annotation crop visibility
+            if (!elevation.get_Parameter(BuiltInParameter.VIEWER_ANNOTATION_CROP_ACTIVE).Set(annotationCrop ? 1 : 0))
+                BH.Engine.Base.Compute.RecordWarning($"Could not set the annotation crop in the view. Parameter is ready-only.");
+
+            return;
+
```
pawelbaran, line 93: redundant

Revit_Core_Engine/Modify/Elevation.cs
```diff
@@ -0,0 +1,124 @@
...
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Modify existing elevation view.")]
```
pawelbaran, line 40: Would be great if you could explain what does the method do - _modify_ can have a very broad meaning

pawelbaran (changes requested):
Sorry for yet another change request @michal-pekacki 🙈

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Input("referenceViewPlan", "ViewPlan that elevation marker is referenced to.")]
+        [Input("depth", "Depth of the elevation view.")]
+        [Input("height", "Height of the elevation view.")]
+        [Input("offset", "Offset that will be added to the elevation view CropBox shape.")]
+        [Input("elevationMarker", "Elevation marker that has been created (out parameter).")]
+        [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
+        [Input("cropRegionVisible", "True if the crop region should be visible.")]
+        [Input("annotationCrop", "True if the annotation crop should be visible.")]
+        [Output("elevationView", "Elevation view that has been created.")]
+        public static ViewSection ElevationByLine(this Line elevationLine, string elevationName, View referenceViewPlan, double depth, double height, double offset, out ElevationMarker elevationMarker, ElementId viewTemplateId = null, bool cropRegionVisible = false, bool annotationCrop = false)
```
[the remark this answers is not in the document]
michal-pekacki: As discussed online - I will remove `ElevationMarker` output from this method.

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
michal-pekacki: I fixed angle calculation 👍

travispotterBH (commented):
In general a big fan of explicit typing (although I am guilty myself of using var at times)

Revit_Core_Engine/Create/View/ElevationMarker.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [Description("Creates elevation marker in the current Revit file.")]
+        [Input("elevationMarkerLocation", "Placement point of the elevation marker.")]
+        [Input("referenceViewPlan", "ViewPlan that elevation marker is referenced to.")]
+        [Output("elevationMarker", "Elevation marker that has been created.")]
+        public static ElevationMarker ElevationMarker(this XYZ elevationMarkerLocation, View referenceViewPlan)
+        {
+            Document doc = referenceViewPlan.Document;
+            var vft = Query.ViewFamilyType(doc, ViewFamily.Elevation);
+            var elevationMarker = Autodesk.Revit.DB.ElevationMarker.CreateElevationMarker(doc, vft.Id, elevationMarkerLocation, referenceViewPlan.Scale);
```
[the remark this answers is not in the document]
michal-pekacki: Done 👍

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -265,7 +267,34 @@ private static XYZ CombineHorizontalWithVerticalAngles(double radiansHorizontalA
...
+            var minZ = bboxPointsTransformed.Select(x => x.Z).OrderBy(x => x).FirstOrDefault();
+            var maxZ = bboxPointsTransformed.Select(x => x.Z).OrderByDescending(x => x).FirstOrDefault();
+
+            var newMin = new XYZ(minX, minY, minZ);
+            var newMax = new XYZ(maxX, maxY, maxZ);
+
+            var newBBox = new BoundingBoxXYZ() { Min = newMin, Max = newMax };
+            view3D.CropBoxActive = true;
+            view3D.CropBox = newBBox;
+            view3D.CropBoxVisible = false;
```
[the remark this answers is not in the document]
michal-pekacki: Done 👍

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check compliance

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -39,23 +41,31 @@ public static partial class Create
...
-            
-            Modify.SetViewDetailLevel(result, viewDetailLevel);
+            newView = Autodesk.Revit.DB.View3D.CreateIsometric(document, vft.Id);
+
+            if (viewDetailLevel != ViewDetailLevel.Undefined)
+                Modify.SetViewDetailLevel(newView, viewDetailLevel);
 
             if (boundingBoxXyz != null)
             {
+                if (offset != 0)
```
pawelbaran, line 59: It is always safer to use some tolerance
```C#
                if (Math.Abs(offset) > tolerance)
```

Revit_Core_Engine/Query/Perimeter.cs
```diff
@@ -44,24 +47,14 @@ public static List<PolyCurve> Perimeter(this SpatialElement spatialElement, Revi
             if (spatialElement == null)
                 return null;
 
-            IList<IList<BoundarySegment>> boundarySegments = spatialElement.GetBoundarySegments(new SpatialElementBoundaryOptions());
-            if (boundarySegments == null)
-                return null;
+            Autodesk.Revit.DB.Face face = spatialElement.Faces(new Options()).Where(x => Math.Abs((x as PlanarFace).FaceNormal.Z + 1) < Tolerance.MacroDistance).FirstOrDefault();
```
pawelbaran, line 50: Why are you changing the methodology of extracting the perimeter here? Is there a bug in the current algorithm? I am asking because we need to be very careful with this method as is it used by multiple convert methods, which means the change will require significant testing and probably agreement from S&P.

Revit_Core_Engine/Create/View/View3D.cs
michal-pekacki: Safety first 👍

Revit_Core_Engine/Create/View/ViewSection/ElevationByLine.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            //check if elevation line is horizontal, project on view plane if not
+            if (lineDirection.DotProduct(XYZ.BasisZ) > Tolerance.Angle)
+            {
+                BH.Engine.Base.Compute.RecordWarning($"The elevation location line is not horizontal and will be projected on reference view plane.");
+                var viewPlane = referenceViewPlan.SketchPlane.GetPlane();
+                var bhomPlane = BH.Engine.Geometry.Create.Plane(viewPlane.Origin.PointFromRevit(), viewPlane.Normal.VectorFromRevit());
+                var bhomProjLine = elevationLine.FromRevit().Project(bhomPlane);
+
+                elevationLine = bhomProjLine.ToRevit();
```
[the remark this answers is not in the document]
michal-pekacki: Done 👍

Revit_Core_Engine/Query/Perimeter.cs
michal-pekacki: I restored this method to previous version.

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -39,23 +41,31 @@ public static partial class Create
...
+                {
+                    boundingBoxXyz.Min -= new XYZ(offset, offset, offset);
+                    boundingBoxXyz.Max += new XYZ(offset, offset, offset);
```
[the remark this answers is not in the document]
michal-pekacki: Updated 👍

pawelbaran (commented):
I am happy with the code changes based on reading the code - good to merge from my perspective as soon as we get tester's approval.

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -39,23 +42,27 @@ public static partial class Create
...
+            View newView = null;
             ViewFamilyType vft = Query.ViewFamilyType(document, ViewFamily.ThreeDimensional);
 
-            result = Autodesk.Revit.DB.View3D.CreateIsometric(document, vft.Id);
-            
-            Modify.SetViewDetailLevel(result, viewDetailLevel);
+            newView = Autodesk.Revit.DB.View3D.CreateIsometric(document, vft.Id);
+
+            if (viewDetailLevel != ViewDetailLevel.Undefined)
+                Modify.SetViewDetailLevel(newView, viewDetailLevel);
```
pawelbaran, line 56: Sorry one more comment: why did you add an `if` statement here? You are effectively culling the warning when the detail level input is undefined, which is undesirable I would say.

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -39,23 +42,27 @@ public static partial class Create
         [Input("viewName", "Optional, name of the new view.")]
         [Input("boundingBoxXyz", "Optional, the cuboid BoundingBoxXYZ to fit isometric view.")]
         [Input("viewTemplateId", "Optional, the View Template Id to be applied in the view.")]
-        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]        
+        [Input("viewDetailLevel", "Optional, the Detail Level of the view.")]
+        [Input("offset", "Offset that will be added to the BoundingBoxXYZ.")]
```
pawelbaran, line 46:
```suggestion
        [Input("offset", "Offset that will be added to each side of the input BoundingBoxXYZ.")]
```

Revit_Core_Engine/Create/View/View3D.cs
```diff
@@ -65,33 +72,28 @@ public static View View3D(this Document document, string viewName = null, Boundi
 
             if (viewTemplateId != null)
             {
+                if (!(document.GetElement(viewTemplateId) as View).IsTemplate)
```
pawelbaran, line 75: This will crash if there is no element under `viewTemplateId`.

```suggestion
                if ((document.GetElement(viewTemplateId) as View)?.IsTemplate != true)
```

...also not sure why did you move this part out of the try/catch clause if you are raising same warning for both cases anyways.

pawelbaran (changes requested):
Last few comments, I also noticed that you have boilerplate around setting view template (the code below being repeated a few times across different view creation method). Wouldn't it be worth centralising this under `SetViewTemplate`, same as we did with `SetViewName`?
image

pawelbaran (changes requested):
Last few comments, I also noticed that you have boilerplate around setting view template (the code below being repeated a few times across different view creation method). Wouldn't it be worth centralising this under `SetViewTemplate`, same as we did with `SetViewName`?
image

Revit_Core_Engine/Create/View/View3D.cs
michal-pekacki: Agreed `if` is not needed here 👍

Revit_Core_Engine/Create/View/View3D.cs
pawelbaran: But isn't it worth dropping a more specific error then? Now you only avoid one `catch`, but that does not change anything else than that, right?

pawelbaran (commented):
Happy with the changes now - waiting for the testers to approve the functionality + the bot checks to pass 👍

michal-pekacki:
@BHoMBot check versioning

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check required

pawelbaran (approved):
Happy to approve following the code review and testing done in collaboration with @michal-pekacki 👍

pawelbaran:
@BHoMBot check ready-to-merge

michal-pekacki:
@BHoMBot check ready-to-merge

PR #1360, opened 2023-06-02
Add `Query.Parameters`

### LINKED ISSUE #1359: Add element parameters method
#### Description:

Add method to return all available element parameters.

michal-pekacki:

Closes #1359 

Method that returns all available and unique parameters of given elements.

### DIFF
--- a/Revit_Core_Engine/Query/Parameters.cs
+++ b/Revit_Core_Engine/Query/Parameters.cs
@@ -36,9 +36,10 @@ namespace BH.Revit.Engine.Core
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Gets the parameter type of a Revit parameter definition.")]
-        [Input("definition", "Revit parameter definition to extract the parameter type from.")]
-        [Output("parameterType", "Parameter type extracted from the input Revit parameter definition.")]
+        [Description("Returns all unique parameters of the elements.")]
+        [Input("elements", "Elements to get the parameters from.")]
+        [Input("includeHiddenParameters", "True for including hidden parameters (not visible in the Revit UI).")]
+        [Output("parameters", "Unique parameters of the given elements.")]
         public static List<Parameter> Parameters(this List<Element> elements, bool includeHiddenParameters = false)
         {
             var parameters = new List<Parameter>();
@@ -53,9 +54,7 @@ namespace BH.Revit.Engine.Core
 
         /***************************************************/
 
-        [Description("Gets the parameter type of a Revit parameter definition.")]
-        [Input("definition", "Revit parameter definition to extract the parameter type from.")]
-        [Output("parameterType", "Parameter type extracted from the input Revit parameter definition.")]
+        [Description("Returns all unique parameters of the element.")]
         private static List<Parameter> Parameters(this Element element, bool includeHiddenParameters = false)
         {
             var doc = element.Document;
--- a/Revit_Core_Engine/Query/Parameters.cs
+++ b/Revit_Core_Engine/Query/Parameters.cs
@@ -39,7 +39,7 @@ namespace BH.Revit.Engine.Core
 
         [Description("Returns single instance of each unique type parameter of the given elements. Each parameter is extracted only from the first element in the group that shares it, therefore no information about values nor parent elements of the parameter objects should be derived from the output of this method.")]
         [Input("elements", "Elements to get the type parameters from.")]
-        [Input("includeHiddenParameters", "True for including hidden parameters (not displayed in the Revit UI).")]
+        [Input("includeHiddenParameters", "If true, include hidden parameters (not displayed in the Revit UI), otherwise skip them.")]
         [Output("parameters", "Unique type parameters of the given elements.")]
         public static List<Parameter> TypeParameters(this List<Element> elements, bool includeHiddenParameters = true)
         {
@@ -51,7 +51,7 @@ namespace BH.Revit.Engine.Core
 
             if (elements.GroupBy(x => x.Document).Count() > 1)
             {
-                BH.Engine.Base.Compute.RecordWarning("Could not return parameters because elements are stored in more than one Revit document.");
+                BH.Engine.Base.Compute.RecordWarning("Could not return parameters because elements are originate from more than one Revit document.");
                 return null;
             }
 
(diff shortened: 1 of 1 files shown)


Revit_Core_Engine/Query/Parameters.cs
```diff
@@ -0,0 +1,78 @@
...
+        [Input("includeHiddenParameters", "True for including hidden parameters (not visible in the Revit UI).")]
+        [Output("parameters", "Unique parameters of the given elements.")]
+        public static List<Parameter> Parameters(this List<Element> elements, bool includeHiddenParameters = false)
+        {
+            var parameters = new List<Parameter>();
+
+            foreach (var elem in elements)
+                parameters.AddRange(elem.Parameters(includeHiddenParameters));
+
+            parameters = parameters.GroupBy(x => x.Id.IntegerValue).Select(x => x.First()).ToList();
```
pawelbaran, line 50: I do not think this will be very efficient, and efficiency may be important here taken we are likely to parse entire documents. I would rather consider creating a `HashSet<int>` where you would store the ElementIds of items already added to the list, so that you would have list of items plus a hash set of their ids to check against (worth having a quick read about performance of `Contains` query in hash-based collections and enumerables btw).

...actually I would be curious how will such tweak affect performance - could you pls compare and drop a screenshot in the comments?

Revit_Core_Engine/Query/Parameters.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            return parameters;
+        }
+
+        /***************************************************/
+
+        [Description("Returns all unique parameters of the element.")]
+        private static List<Parameter> Parameters(this Element element, bool includeHiddenParameters = false)
```
pawelbaran, line 58: I would make this method public

Revit_Core_Engine/Query/Parameters.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns all unique parameters of the elements.")]
+        [Input("elements", "Elements to get the parameters from.")]
+        [Input("includeHiddenParameters", "True for including hidden parameters (not visible in the Revit UI).")]
+        [Output("parameters", "Unique parameters of the given elements.")]
+        public static List<Parameter> Parameters(this List<Element> elements, bool includeHiddenParameters = false)
```
pawelbaran, line 43: Is there a rationale behind `includeHiddenParameters = false`? My gut feeling would be to make it `true` by default, taken in most contexts the users will rather not want to see these parameters. But happy to be convinced otherwise 👍

pawelbaran (changes requested):
I would add one more method:
```C#
public static List<Parameter> TypeParameters(this List<Element> elements, bool includeHiddenParameters = false)
{
    // 1. Get all unique ids of element types (element.GetTypeId())
    // 2. Call and return types.Parameters(includeHiddenParameters)
}
```

Also I would rather prefer following the general BHoM convention and using explicit typing, following the recent discussions that we had.

pawelbaran (changes requested):
I would add one more method:
```C#
public static List<Parameter> TypeParameters(this List<Element> elements, bool includeHiddenParameters = false)
{
    // 1. Get all unique ids of element types (element.GetTypeId())
    // 2. Call and return types.Parameters(includeHiddenParameters)
}
```

Also I would rather prefer following the general BHoM convention and using explicit typing, following the recent discussions that we had.

Revit_Core_Engine/Query/Parameters.cs
michal-pekacki: Will do, as discussed and tested offline, performance is better for larger number of elements.

Revit_Core_Engine/Query/Parameters.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Query/Parameters.cs
michal-pekacki: Agree and will change it 👍

Revit_Core_Engine/Query/Parameters.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            if (elements == null)
+                return null;
+
+            if (elements.Count == 0)
+                return new List<Parameter>();
+
+            if (elements.GroupBy(x => x.Document).Count() > 1)
+            {
+                BH.Engine.Base.Compute.RecordWarning("Could not return parameters because elements are stored in more than one Revit document.");
```
pawelbaran, line 54:
```suggestion
                BH.Engine.Base.Compute.RecordWarning("Could not return parameters because elements originate from more than one Revit document.");
```

Revit_Core_Engine/Query/Parameters.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns single instance of each unique type parameter of the given elements. Each parameter is extracted only from the first element in the group that shares it, therefore no information about values nor parent elements of the parameter objects should be derived from the output of this method.")]
+        [Input("elements", "Elements to get the type parameters from.")]
+        [Input("includeHiddenParameters", "True for including hidden parameters (not displayed in the Revit UI).")]
```
pawelbaran, line 42: Minor alignment with other descriptions of `bool` method params recommended (refers to same input in other two methods)

```suggestion
        [Input("includeHiddenParameters", "If true, include hidden parameters (not displayed in the Revit UI), otherwise skip them.")]
```

pawelbaran (commented):
Last minor bits 😉

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check required

pawelbaran (approved):
Happy to approve following a quality dev-review process - thanks @michal-pekacki 👍

pawelbaran:
@BHoMBot check ready-to-merge

PR #1380, opened 2023-06-23
Add new methods to Revit_Core_Engine

### LINKED ISSUE #1343: Additional methods for Revit_Core_Engine
#### Description:

To use Geometry_Engine methods, we often need to convert Revit native geometry elements to Bhom objects. This becomes inefficient in heavy applications. Therefore, we need a few additional geometry methods in Revit_Core_Engine to handle use cases like:
- Get the intersection points of 2 vectors
- Get the smallest possible angle between 2 vectors
- Get the closest point from a point cloud
- Check if a line or face is vertical

vietle-bh:

Closes #1343 

Added new methods such as:
- Get the intersection points of 2 vectors
- Get the smallest possible angle between 2 vectors
- Get the closest point from a point cloud
- Check if a line or face is vertical


vietle-bh:
Sorry @enarhi , I requested your review by mistake 😅 
Have a nice weekend!

PR #1415, opened 2023-10-09
Improve `SpatialBoundCondition` values

### LINKED ISSUE #1414: Improve values of SpatialBoundCondition
#### Description:

`Revit_Core_Engine\Query\SpatialBoundCondition.cs` currently holds 4 possible bounding statuses for a space: Bounded, Unplaced, NotEnclosed & Redundant. However, Redundant implies the space is no longer needed, although spaces get this status because they overlap with other spaces in the same volume. This can be confusing, especially when we reflect these statuses into any kind of log to report the output of a method. I suggest renaming Redundant to Overlapping.

Also, there's 1 more status missing from the current enum. Some spaces that are too complex for Revit to determine the bounding condition will raise the warning below and disappear from their views:

image

We should add a new `NoGeometry` enum value to cover this case. The Query method for `SpatialBoundCondition` also needs an update to handle this edge case.

vietle-bh:

Closes #1414 

- Renamed Redundant to Overlapping
- Added NoGeometry
- Added conditions for arriving at NoGeometry to the `SpatialBoundCondition` query method.

Please note the edge case for NoGeometry only exists in Revit models with volume computation turned on:

image

### Test files

Building the repo should raise no error, and bot checks should pass.

### DIFF
--- a/Revit_Core_Engine/Query/SpatialBoundCondition.cs
+++ b/Revit_Core_Engine/Query/SpatialBoundCondition.cs
@@ -51,7 +51,7 @@ namespace BH.Revit.Engine.Core
                 }
                 else if (elem.GetBoundarySegments(new SpatialElementBoundaryOptions()).Count > 0)
                 {
-                    return BoundCondition.Redundant;
+                    return BoundCondition.Overlapping;
                 }
                 else
                 {
--- a/Revit_Core_Engine/Query/SpatialBoundCondition.cs
+++ b/Revit_Core_Engine/Query/SpatialBoundCondition.cs
@@ -38,7 +38,7 @@ namespace BH.Revit.Engine.Core
         [Description("Check if a room/space/area is bounded, unplaced, redundant, or not enclosed.")]
         [Input("elem", "A spatial element to be check if it's bounded, unplaced, redundant, or not enclosed.")]
         [Output("BoundCondition", "The bound condition of the spatial element to be checked.")]
-        public static BoundCondition? SpatialBoundCondition(this SpatialElement elem, bool volumeComputationEnabled = false)
+        public static BoundCondition? SpatialBoundCondition(this SpatialElement elem)
         {
             if (elem == null)
                 return null;
@@ -58,7 +58,7 @@ namespace BH.Revit.Engine.Core
                     return BoundCondition.NotEnclosed;
                 }
             }
-            else if (volumeComputationEnabled)
+            else if (AreaVolumeSettings.GetAreaVolumeSettings(elem.Document).ComputeVolumes)
             {
                 if (elem is Room room && room.Volume == 0
                     || elem is Space space && space.Volume == 0)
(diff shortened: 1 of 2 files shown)


vietle-bh:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check null-handling
@BHoMBot check serialisation
@BHoMBot check project-compliance

vietle-bh:
@BHoMBot check versioning 
@BHoMBot check installer

vietle-bh:
@BHoMBot check versioning

Revit_oM/Enums/BoundCondition.cs
```diff
@@ -30,14 +30,15 @@ namespace BH.oM.Adapters.Revit.Enums
         "Below is the meaning of each condition for rooms. The same applies to spaces and areas." +
         "- Unplaced: https://help.autodesk.com/view/RVT/2022/ENU/?guid=GUID-BC1DC181-B6D0-4479-8385-363A9EE5E75E" +
         "- Not enclosed: https://help.autodesk.com/view/RVT/2022/ENU/?guid=GUID-1AEDF540-7CB3-4CAB-885A-ACDF70154312" +
-        "- Redundant: https://help.autodesk.com/view/RVT/2022/ENU/?guid=GUID-0DF409DD-3BBD-4488-B544-D075D1807747" +
+        "- Redundant(Overlapping another space): https://help.autodesk.com/view/RVT/2022/ENU/?guid=GUID-0DF409DD-3BBD-4488-B544-D075D1807747" +
```
pawelbaran, line 33: Shouldn't this line start from 'Overlapping:' to follow the convention? Also missing description for no geometry

Revit_Core_Engine/Query/SpatialBoundCondition.cs
```diff
@@ -51,13 +51,23 @@ public static partial class Query
...
+                    return BoundCondition.Overlapping;
                 }
                 else
                 {
                     return BoundCondition.NotEnclosed;
                 }
             }
+            else if (AreaVolumeSettings.GetAreaVolumeSettings(elem.Document).ComputeVolumes)
+            {
+                if (elem is Room room && room.Volume == 0
```
pawelbaran, line 63: Wouldn't it be leaving some tolerance, e.g. `room.Volume < 1e-3` or similar?

pawelbaran (commented):
Looks good overall, just two questions/comments left

Revit_Core_Engine/Query/SpatialBoundCondition.cs
vietle-bh: Actually, the volume will always be exactly 0 when the spatial element has no geometry, so I think this is safe enough. This is similar to a space that has been unplaced (deleted from the plan view) which has an area of exactly 0.

vietle-bh:
@BHoMBot check required

pawelbaran (dismissed):
Changes and responses to my previous comments make full sense, happy to approve 👍

pawelbaran:
@BHoMBot check ready-to-merge

vietle-bh:
@BHoMBot check versioning

vietle-bh:
@BHoMBot check installer

pawelbaran (approved):
Happy to approve from the code side, functionality to be tested on alpha 👍

pawelbaran:
@BHoMBot check ready-to-merge

PR #1431, opened 2023-12-29
'Edited by' added to the default blacklist in diffing

### LINKED ISSUE #1430: Diffing: should 'Edited by' parameter be ignored by default?
#### Description:
While doing various diffing exercises, I noticed we keep on ruling out `Edited by` parameter - maybe we should simply add it to the ignored set by default? Thoughts against?

@alelom FYI the parameter informs about the last user that 'touched' a Revit element. So as long as it may be an important piece of information in many contexts, we usually do not care about it in diffing.

pawelbaran:

Closes #1430

### Test files

Standard diffing test procedure

### Additional comments

...I thought I will PR that tweak as I already had it ready on my local - happy to put the PR down if we end up disagreeing with the idea 😉

### DIFF
--- a/Revit_oM/Config/RevitComparisonConfig.cs
+++ b/Revit_oM/Config/RevitComparisonConfig.cs
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2023, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2024, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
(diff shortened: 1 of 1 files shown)


vietle-bh (approved):
I support this change 👍 
Let's see if @alelom feels the same 😄

alelom (approved):
Makes total sense to me! Approving on the base of code changes, not tested.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check required

FraserGreenroyd (approved):
Reapproving based on previous reviews from @vietle-bh and @alelom as only change has been to copyright compliance since.

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #1442, opened 2024-01-26
View queries added

### LINKED ISSUE #1440: Add ability to query views for their depth ranges, bounding solids, bboxes etc.
#### Description:

As in the title, we often work with filtering of elements against views - would be great to have a suite of methods that would return their defining range, solid, bounding box etc.

I have the code scattered across different projects, will try to capture it under a single PR.

pawelbaran:

Closes #1440

### DIFF
--- a/Revit_Core_Engine/Create/CurveLoop.cs
+++ b/Revit_Core_Engine/Create/CurveLoop.cs
@@ -1,0 +1,59 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2024, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using BH.oM.Base.Attributes;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Creates a CurveLoop representing a polyline defined by a given collection of points.")]
+        [Input("points", "Collection of points defining the CurveLoop to be created.")]
+        [Input("close", "If true, the method will ensure the returned CurveLoop is closed.")]
+        [Output("loop", "CurveLoop representing a polyline defined by the input collection of points.")]
+        public static CurveLoop CurveLoop(List<XYZ> points, bool close = true)
+        {
+            if (points == null)
+                return null;
+
+            CurveLoop cl = new CurveLoop();
+            for (int i = 1; i < points.Count; i++)
+            {
+                cl.Append(Line.CreateBound(points[i - 1], points[i]));
+            }
+
+            if (close)
+                cl.Append(Line.CreateBound(points[points.Count - 1], points[0]));
+
+            return cl;
+        }
+
+        /***************************************************/
+    }
+}
--- a/Revit_Core_Engine/Query/ViewSolid.cs
+++ b/Revit_Core_Engine/Query/ViewSolid.cs
@@ -72,7 +72,7 @@ namespace BH.Revit.Engine.Core
                 if (loops == null || loops.Count == 0)
                     return null;
 
-                Line range = view.ViewDepth();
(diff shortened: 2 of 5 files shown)


pawelbaran:
@BHoMBot check compliance

Revit_Core_Engine/Create/CurveLoop.cs
```diff
@@ -0,0 +1,59 @@
...
+            for (int i = 1; i < points.Count; i++)
+            {
+                cl.Append(Line.CreateBound(points[i - 1], points[i]));
+            }
+
+            if (close)
+                cl.Append(Line.CreateBound(points[points.Count - 1], points[0]));
+
+            return cl;
+        }
```
travispotterBH, line 55:
```suggestion
        [Description("Creates an open CurveLoop representing a polyline defined by a given collection of points.")]
        [Input("points", "Collection of points defining the CurveLoop to be created.")]
        [Output("loop", "CurveLoop representing a polyline defined by the input collection of points.")]
        public static CurveLoop OpenCurveLoop(List<XYZ> points)
        {
            if (points == null)
                return null;

            CurveLoop cl = new CurveLoop();
            for (int i = 1; i < points.Count; i++)
            {
                cl.Append(Line.CreateBound(points[i - 1], points[i]));
            }
            return cl;
        }

        [Description("Creates a closed CurveLoop representing a polyline defined by a given collection of points.")]
        [Input("points", "Collection of points defining the CurveLoop to be created.")]
        [Output("loop", "CurveLoop representing a polyline defined by the input collection of points.")]
        public static CurveLoop ClosedCurveLoop(List<XYZ> points)
        {
               CurveLoop curveLoop = OpenCurveLoop(points);
                curveLoop.Append(Line.CreateBound(points[points.Count - 1], points[0]));

            return curveLoop;
        }
```

suggestion (DevEx): If we specify in the method name that the return of Open vs Closed type of CurveLoop, it can be a bit easier to reason through the code on a first pass rather than intuiting what kind of curve loop you have based on a boolean flag in the method signature.

example:

```csharp

bool close = // some condition that makes this closed.
  var curveLoop = close ? ClosedCurveLoop(points) : OpenCurveLoop(points);

```
versus 

```csharp

//without hovering over the method signature or reading the documentation, it is difficult to know why you need the boolean, or what the state of the returned object is.
var openCurveLoop = CurveLoop(points, false);
var closedCurveLoop = CurveLoop(points, true);

```

pawelbaran: Personally I would disagree. This sort of verbosity makes the code/API less readable to me, and goes against the most common pattern - I have never seen a polyline creation methods to be split to separate signatures per open/closed case.

FraserGreenroyd: Thought: Given this is Revit_Toolkit and sits at the heart of all Revit Development we do within the BHoM community (which granted, currently, isn't very much beyond this toolkit, but we have plans for it to grow into zero code tools, etc.), and we expect this toolkit to serve as an API as much as a usable toolkit for other tools to build upon, I think verbosity can be beneficial in this case.

As @travispotterBH highlights, it makes the use of each method very clear to the calling code, and reduces the risk of someone putting the wrong boolean flag and getting the wrong result. It reduces cognitive overload during debugging by saving us needing to refer to documentation (which currently isn't as accessible to developers as it is to users) for what the boolean does.

I agree that in BHoM land currently we don't have many examples of this practice, our polylines for example are defined as closed if the first and last points are identical (duplicating data) vs Rhino who, I believe (please correct me if I'm wrong, it's been a while since I looked at the Rhino API) has a boolean flag property on the polyline determining whether its closed or not. However, by way of offering a 'similar but different' option, [this is a method for not closed spaces](https://github.com/BHoM/BHoM_Engine/blob/develop/Environment_Engine/Query/NotClosedSpaces.cs#L50) and [this is a method for closed spaces](https://github.com/BHoM/BHoM_Engine/blob/develop/Environment_Engine/Query/ClosedSpaces.cs#L50) within the Environment Engine.

So my two cents there, a bit of extra verbosity in an API level toolkit has some potential benefit down the line.

Suggestion: Use `points.Last()` instead of `points[points.Count - 1]` for the last point? 😋

pawelbaran: From what I understand, the methods you shared are quite different from this case: the methods are querying a set of spaces for open or closed ones, while here we only create a polyline. Extraction vs creation. The usual convention for creating a polyline is to add `closed` as a boolean flag at the end of the method signature and I believe any developer will look for it there rather than in separate methods.

I believe we may need a wider discussion about verbosity of method signatures if such case leads us to waste a combined hour of time (or whatever the number will be at the end of this thread). But please let's not start it now...

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -0,0 +1,113 @@
...
+                loops = pts.Select(x => Create.CurveLoop(x)).ToList();
+                return GeometryCreationUtilities.CreateExtrusionGeometry(loops, range.Direction, range.Length);
+            }
+            else
+                return null;
+        }
+
+        /***************************************************/
+
+        private static Line ViewDepth(this View view)
```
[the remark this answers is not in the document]
pawelbaran: I would not know what does `ViewBoundAxis` means either, but I renamed the method to `ViewDepthLine` - hopefully you like it too 😉

pawelbaran:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check null-handling
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check versioning
@BHoMBot check installer

pawelbaran:
@BHoMBot check required

vietle-bh (approved):
Tested via https://github.com/BuroHappoldEngineering/Revit_Tools_Test/pull/37

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 21163387754

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #1453, opened 2024-02-12
Fix ordering of surface boundary loops

### LINKED ISSUE #1452: Wrong curve loop order in surface conversion from Revit
#### Description:

The current method for converting a Revit surface to Bhom geometry assumes that the first curve loop from the Revit face is the main boundary loop. This isn't always the case.

https://github.com/BHoM/Revit_Toolkit/blob/0057b3dc28d52730b71d1fb7c4c122b8eb4188b6/Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs#L46-L48

#### Expected behaviour:

We need to check for the max perimeter to accurately determine the main boundary loop.

vietle-bh:

Closes #1452 

The proposed approach calculates how many points in each curve loop actually sit on the surface's convex hull. It then orders boundary loops descendingly by these numbers to ensure the first item will always be the main external loop.

### Test files

Given the PR's scope, probably just code review is enough? 😄


vietle-bh:
@BHoMBot check required

vietle-bh:

pawelbaran:
Capturing offline conversations with @vietle-bh to keep transparent:
- let's provide a test file to work against
- good to consider two alternatives:
    - checking if inner/outer outlines are not grouped by direction (clockwise/counterclockwise) - pretty common in geometry
    - sorting by area - largest will be the outer outline, all others inner

vietle-bh:
@pawelbaran : Using winding direction works. Thanks! Please re-review 👍

Revit_Core_Engine/Modify/SortBoundaryLoops.cs
```diff
@@ -0,0 +1,78 @@
...
+        [Description("Sort a list of curve loops (from a spatial element) so that the first one corresponds to the main loop that contains all others.")]
+        [Input("loops", "Boundary curve loops from a spatial element like a room or space.")]
+        [Output("normal", "Normal vector of the surface the curve loops come from, usually come from face.ComputeNormal or face.FaceNormal if it's a planar face.")]
+        public static void SortBoundaryLoops(this List<CurveLoop> loops, XYZ normal)
+        {
+            if (loops.Count < 2)
+                return;
+
+            //By convention, the main boundary must be counter-clockwise relative to the normal vector.
+            while (loops[0].IsCounterclockwise(normal) == false)
```
pawelbaran, line 45: Couldn't we find the index of first counterclockwise curve in set with `FindIndex`, then remove at that index and insert the curve at 0?

pawelbaran (commented):
Happy to see this being picked up @vietle-bh 👍 I have one question, please see the code comment

Revit_Core_Engine/Query/SolidFilledRegionType.cs
```diff
@@ -0,0 +1,59 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Returns the 'Solid' filled region type which exists in all Revit models.")]
+        [Input("doc", "The active Revit document.")]
+        [Output("filledRegionType", "The 'Solid' filled region type which exists in all Revit models.")]
+        public static FilledRegionType SolidFilledRegionType(this Document doc)
```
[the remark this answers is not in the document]
vietle-bh: It feels like either a solid background or solid foreground should be good enough to call a pattern "solid". Anyway, I'll remove this method as it's only needed for the test method of SortBoundaryLoops, and now SortBoundaryLoops will be gone so we can't test it anymore.

vietle-bh:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check null-handling
@BHoMBot check serialisation
@BHoMBot check project-compliance

Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs
```diff
@@ -43,9 +43,9 @@ public static oM.Geometry.PlanarSurface FromRevit(this PlanarFace face)
             if (face == null)
                 return null;
 
-            IList<CurveLoop> crvLoop = face.GetEdgesAsCurveLoops();
-            oM.Geometry.ICurve externalBoundary = crvLoop[0].FromRevit();
-            List<oM.Geometry.ICurve> internalBoundary = crvLoop.Skip(1).Select(x => x.FromRevit() as oM.Geometry.ICurve).ToList();                           
+            List<CurveLoop> crvLoops = face.GetEdgesAsCurveLoops().ToList();
+            oM.Geometry.ICurve externalBoundary = crvLoops.FirstOrDefault(x => x.IsCounterclockwise(face.FaceNormal))?.FromRevit();
+            List<oM.Geometry.ICurve> internalBoundary = crvLoops.Where(x => x != externalBoundary).Select(x => x.FromRevit() as oM.Geometry.ICurve).ToList();
```
pawelbaran, line 48:
```suggestion
            CurveLoop outer = crvLoops.FirstOrDefault(x => x.IsCounterclockwise(face.FaceNormal));
oM.Geometry.ICurve externalBoundary = outer?.FromRevit();
            List<oM.Geometry.ICurve> internalBoundary = crvLoops.Where(x => x != outer).Select(x => x.FromRevit() as oM.Geometry.ICurve).ToList();
```

pawelbaran (changes requested):
I must have sounded very confident so that you took my suggestion without critique 😉 nothing more fun than bugfixing own self!

BTW, this PR requires a rebase I'm afraid

Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs
vietle-bh: Now I'm hyper-critical of anything you wrote! 🤣 
How about this?

```c#
        public static oM.Geometry.PlanarSurface FromRevit(this PlanarFace face)
        {
            if (face == null)
                return null;

            List<CurveLoop> crvLoops = face.GetEdgesAsCurveLoops().ToList();
            CurveLoop externalLoop = crvLoops.FirstOrDefault(x => x.IsCounterclockwise(face.FaceNormal));

            if (externalLoop == null)
            {
                BH.Engine.Base.Compute.RecordError($"A Revit face has no counter-clockwise boundary loop. Its conversion to BHoM ICurve may be inaccurate.");
                externalLoop = crvLoops[0];
            }

            oM.Geometry.ICurve externalBoundary = externalLoop.FromRevit();
            List<oM.Geometry.ICurve> internalBoundaries = crvLoops.Where(x => x != externalLoop).Select(x => x.FromRevit() as oM.Geometry.ICurve).ToList();

            return new oM.Geometry.PlanarSurface(externalBoundary, internalBoundaries);
        }
```

pawelbaran: I think I would return `null` in case of external loop not found - better to return nothing than a surface with wrong outline (I think the code would break further downstream anyways). So I would opt for something along the lines:
```C#
if (externalLoop == null)
{
    BH.Engine.Base.Compute.RecordError($"Conversion of element face to planar surface failed because face outline could not be found.");
    return null;
}
```

pawelbaran (approved):
Happy to approve now, nice scope limitation finally @vietle-bh 👍

pawelbaran:
@BHoMBot check required

vietle-bh:
@BHoMBot check ready-to-merge

PR #1461, opened 2024-03-18
Parameter From Revit - parameter value bug fix

### LINKED ISSUE #1460: Parameter From Revit - piping system parameter value bug
#### Description:

Parameter From Revit method crashes on Piping System `Volume` parameter.

michal-pekacki:

Closes #1460 

Added try/catch to querying parameter value to avoid method crashes.


pawelbaran:
@FraserGreenroyd FYI this PR does not go into `develop` before Beta release, it is an edge case spotted while extremely heavy testing (probably not possible to reproduce manually)

michal-pekacki:
Bug has been fixed outside Revit_Toolkit repo.

PR #1468, opened 2024-04-02
Triangulation factor for Medium mesh quality changed from 0.5 to 0.3

### LINKED ISSUE #1439: 'Medium' mesh quality
#### Description:
Revit operates with 3 view qualities: `Coarse`, `Medium` and `Fine`. While meshing, these enum values get converted to a mesh quality factor that needs to be in range between 1 and is passed to the triangulation function. When writing the original code I took the most obvious values off the top of my head, being:
- `Coarse` => 0.0
- `Medium` => 0.5
- `Fine` => 1.0

These work fine in general, but I recently ran into performance-related issues caused by the `Medium` value, being a default one in the diffing processes (in general we do not want to use neither of the remaining two, one being too coarse the other too fine). I started investigating how does the mesh quality factor affect mesh size and visual quality. I ran a sensitivity study using a set of Revit elements of different geometrical complexity, from simple cuboids to doubly curved surfaces. I meshed them with different mesh quality factors between 0.0 and 0.5, with the following results:

Top row - vertex count
Bottom row - face count
image

It can clearly be seen that the mesh size and performance starts to grow quickly for mesh quality factors >0.3, which means that we would ideally use a value of 0.3 as `Medium`. I decided to verify whether this would not affect the visual experience of the users, results below.

Visual quality - 0.0
image

Visual quality - 0.1
image

Visual quality - 0.2
image

Visual quality - 0.3
image

Visual quality - 0.4
image

Visual quality - 0.5
image

In my opinion the difference in visual experience between 0.3 and 0.5 is not worth the performance hit, so I would consider changing the mesh quality factor for `Medium` quality from 0.5 to 0.3. This, however, would affect the existing diffing workflows because all elements would flag as geometry changed. This of course is a breaking change, which would require careful treatment, i.e. a piece of information to be embedded in the diffing workflows that would inform the user about incompatibility between 7.0 and older extractions and 7.1 and above.

I believe it is worth the game, but wanted to check with you:
- do you agree the mesh quality achieved with factor 0.3 is visually acceptable as default preview?
- do you agree it makes sense to make an effort and introduce the breaking change to shrink the extreme case meshes by >90%?

For these interested, the sensitivity study is available [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FRevit%5FToolkit%2F%231439%2DMediumMeshQuality&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

pawelbaran:

Closes #1439

### Additional comments

Sensitivity study and discussion available under #1439

### DIFF
--- a/Revit_Core_Engine/Query/FaceTriangulationFactor.cs
+++ b/Revit_Core_Engine/Query/FaceTriangulationFactor.cs
@@ -44,7 +44,7 @@ namespace BH.Revit.Engine.Core
                 case Autodesk.Revit.DB.ViewDetailLevel.Fine:
                     return 1;
                 default:
-                    return 0.5;
+                    return 0.3;
             }
         }
 


FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 23337294228

PR #1495, opened 2024-07-25
Update `LinkedElementsInView` method

### LINKED ISSUE #1494: Update `LinkedElementsInView`
#### Description:

Currently `LinkedElementsInView` works only for `ViewPlan`. The same method working for `View` should be added.

michal-pekacki:

Closes #1494 

Added `LinkedElementsInView` that works for any `View`.

### DIFF
--- a/Revit_Core_Engine/Query/LinkedElementsInView.cs
+++ b/Revit_Core_Engine/Query/LinkedElementsInView.cs
@@ -34,8 +34,8 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.LinkedElementsInView(Autodesk.Revit.DB.ViewPlan, Autodesk.Revit.DB.RevitLinkInstance, System.Collections.Generic.List<Autodesk.Revit.DB.ElementFilter>)")]
-        [Description("Return elements from the revit link instance located in the view scope.")]
-        [Input("view", "View to get visible elements from.")]
+        [Description("Return elements from the revit link instance located in the view scope (including hidden elements)")]
+        [Input("view", "View to get visible elements from. The view needs to belong to the host document.")]
         [Input("linkInstance", "Revit link instance to get the elements from.")]
         [Input("elementFilters", "Additional filters for the element collector. If null, no additional filters will be applied.")]
         [Output("elements", "Elements of link instance in given view.")]
@@ -61,6 +61,8 @@ namespace BH.Revit.Engine.Core
             return elementIds;
         }
 
+        /***************************************************/
+        /****              Private methods              ****/
         /***************************************************/
 
         [Description("Return view solid transformed by inverse link instnace transform.")]
--- a/Revit_Core_Engine/Query/LinkedElementsInView.cs
+++ b/Revit_Core_Engine/Query/LinkedElementsInView.cs
@@ -34,7 +34,7 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.LinkedElementsInView(Autodesk.Revit.DB.ViewPlan, Autodesk.Revit.DB.RevitLinkInstance, System.Collections.Generic.List<Autodesk.Revit.DB.ElementFilter>)")]
-        [Description("Return elements from the revit link instance visible in the view.")]
+        [Description("Return elements from the revit link instance located in the view scope.")]
         [Input("view", "View to get visible elements from.")]
         [Input("linkInstance", "Revit link instance to get the elements from.")]
         [Input("elementFilters", "Additional filters for the element collector. If null, no additional filters will be applied.")]
--- a/Revit_Core_Engine/Query/LinkedElementsInView.cs
+++ b/Revit_Core_Engine/Query/LinkedElementsInView.cs
@@ -33,6 +33,7 @@ namespace BH.Revit.Engine.Core
         /****              Public methods               ****/
         /***************************************************/
 
+        [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.LinkedElementsInView(Autodesk.Revit.DB.ViewPlan, Autodesk.Revit.DB.RevitLinkInstance, System.Collections.Generic.List<Autodesk.Revit.DB.ElementFilter>)")]
         [Description("Return elements from the revit link instance visible in the view.")]
         [Input("view", "View to get visible elements from.")]
         [Input("linkInstance", "Revit link instance to get the elements from.")]
--- a/Revit_Core_Engine/Query/LinkedElementsInView.cs
+++ b/Revit_Core_Engine/Query/LinkedElementsInView.cs
@@ -21,6 +21,7 @@
  */
 
 using Autodesk.Revit.DB;
+using BH.oM.Base;
 using BH.oM.Base.Attributes;
 using System.Collections.Generic;
 using System.ComponentModel;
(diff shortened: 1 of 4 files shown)


Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -87,7 +86,45 @@ private static Solid CropBoxSolid(this View view)
...
+
+        private static Solid NoCropBoxSolid(this View view)
+        {
+            if (view is ViewPlan viewplan)
+            {
+                Output<double, double> range = viewplan.PlanViewRange();
+                BoundingBoxXYZ viewBBox = new BoundingBoxXYZ
+                {
+                    Min = new XYZ(-1e6, -1e6, range.Item1),
+                    Max = new XYZ(1e6, 1e6, range.Item2)
```
pawelbaran, line 107: Could we please introduce a local variable `private static double m_DefaultHorizontalExtents = 1e+6;`?

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -87,7 +86,45 @@ private static Solid CropBoxSolid(this View view)
                 return GeometryCreationUtilities.CreateExtrusionGeometry(loops, range.Direction, range.Length);
             }
             else
+            {
+                if (createIfNoCropBox)
```
pawelbaran, line 90: I would rather rephrase these 3 lines into one:

```C#
return createIfNoCropBox ? NoCropBoxSolid(view) : null
```

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -38,12 +38,12 @@ public static partial class Query
         [Description("Returns a solid that represents the 3-dimensional extents of a given view.")]
         [Input("view", "View to compute the solids.")]
         [Output("solid", "Solid that represents the 3-dimensional extents of the input view.")]
-        public static Solid ViewSolid(this View view)
+        public static Solid ViewSolid(this View view, bool createIfNoCropBox = false)
```
pawelbaran, line 41: Maybe `createIfUncropped`? Also a good description that 1e+6 in directions perpendicular to view directions will be created etc.

Revit_Core_Engine/Query/ViewSolid.cs
michal-pekacki: I used existing `m_DefaultExtents` that is in used inside `ViewDepthLine` method. I think we should centralize this value.

Revit_Core_Engine/Query/ViewSolid.cs
michal-pekacki: Done 👍

Revit_Core_Engine/Query/ViewSolid.cs
michal-pekacki: Please ignore previous comment, I added `m_DefaultHorizontalExtents` as suggested 👍

Revit_Core_Engine/Query/LinkedElementsInView.cs
```diff
@@ -34,29 +33,46 @@ public static partial class Query
         /****              Public methods               ****/
         /***************************************************/
 
-        [Description("Return elements from link instance located in given view. Method returns all instance elements in the view scope (including hidden elements).")]
-        [Input("view", "ViewPlan to get the elements from.")]
-        [Input("linkInstance", "Revit Link Instance to get the elements from.")]
+        [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.LinkedElementsInView(Autodesk.Revit.DB.ViewPlan, Autodesk.Revit.DB.RevitLinkInstance, System.Collections.Generic.List<Autodesk.Revit.DB.ElementFilter>)")]
+        [Description("Return elements from the revit link instance located in the view scope.")]
```
pawelbaran, line 37: Does the method return the hidden elements too? If so, I would consider coming back to the description as in `develop`.

Also worth explaining that the view needs to belong to the host document

Revit_Core_Engine/Query/LinkedElementsInView.cs
```diff
@@ -34,29 +33,46 @@ public static partial class Query
...
+        [Input("view", "View to get visible elements from.")]
```
pawelbaran, line 38:
```suggestion
        [Input("view", "View to get visible elements from. The view needs to belong to the host document.")]
```

Revit_Core_Engine/Query/LinkedElementsInView.cs
```diff
@@ -34,29 +33,46 @@ public static partial class Query
...
+                LogicalAndFilter elementSolidFilter = new LogicalAndFilter(elementFilter, solidFilter);
+                elementIds = new FilteredElementCollector(linkDoc).WherePasses(elementSolidFilter).WhereElementIsNotElementType().ToElementIds();
+            }
+
+            return elementIds;
+        }
 
-            LogicalOrFilter elementFilter = new LogicalOrFilter(elementFilters);
-            LogicalAndFilter elementSolidFilter = new LogicalAndFilter(elementFilter, solidFilter);
+        /***************************************************/
```
pawelbaran, line 64: Private methods header

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -35,15 +35,17 @@ public static partial class Query
         /****              Public methods               ****/
         /***************************************************/
 
+        [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.ViewSolid(Autodesk.Revit.DB.View)")]
         [Description("Returns a solid that represents the 3-dimensional extents of a given view.")]
         [Input("view", "View to compute the solids.")]
+        [Input("createIfUncropped", "Create solid if crop view property is inactive. New solid will be extended by maximum extends value in directions perpendicular to the view type.")]
         [Output("solid", "Solid that represents the 3-dimensional extents of the input view.")]
-        public static Solid ViewSolid(this View view)
+        public static Solid ViewSolid(this View view, bool createIfUncropped = false)
```
pawelbaran, line 43: Now looking at the method signature, I would consider being even more explicit in parameter naming - the more informative we are, the easier to use the method without digging in descriptions 👍 
```suggestion
        public static Solid ViewSolid(this View view, bool createUnlimitedIfViewUncropped = false)
```

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -35,15 +35,17 @@ public static partial class Query
         /****              Public methods               ****/
         /***************************************************/
 
+        [PreviousVersion("7.3", "BH.Revit.Engine.Core.Query.ViewSolid(Autodesk.Revit.DB.View)")]
         [Description("Returns a solid that represents the 3-dimensional extents of a given view.")]
         [Input("view", "View to compute the solids.")]
+        [Input("createIfUncropped", "Create solid if crop view property is inactive. New solid will be extended by maximum extends value in directions perpendicular to the view type.")]
```
pawelbaran, line 41:
```suggestion
        [Input("createIfUncropped", "If false, the method will return null in case of missing crop box (i.e. the view is unlimited, so it cannot be represented by a solid). If true, uncropped views will produce an 'unlimited' solid roughly 1e+6 by 1e+6 in dimensions perpendicular to the view direction, with depth equal to view depth or 1e+4 in case of views without depth.")]
```

Revit_Core_Engine/Query/ViewSolid.cs
```diff
@@ -115,6 +165,12 @@ private static Line ViewDepthLine(this View view)
             }
         }
 
+        /***************************************************/
+        /****              Private fields               ****/
+        /***************************************************/
+
+        private static double m_DefaultHorizontalExtents = 1e+6;
```
pawelbaran, line 172: Could we please put this field next to `m_DefaultExtents` and rename the latter to `m_DefaultVerticalExtents`?

Revit_Core_Engine/Query/LinkedElementsInView.cs
michal-pekacki: Added 👍

Revit_Core_Engine/Query/LinkedElementsInView.cs
michal-pekacki: Added 👍

Revit_Core_Engine/Query/ViewSolid.cs
michal-pekacki: Done 👍

pawelbaran (approved):
Approving on the basis of code review and testing

pawelbaran:
@BHoMBot check compliance

michal-pekacki:
@BHoMBot check required

michal-pekacki:
@BHoMBot check ready-to-merge

PR #1500, opened 2024-08-15
Fix PlanarSurface conversion bug

### LINKED ISSUE #1499: PlanarFace to PlanarSurface conversion assigns the same curve loop twice
#### Description:

I noticed that the conversion of some surfaces to Bhom assigned the same curve loop as both the external boundary and one of the internal openings. Looks like each time we call `face.ExternalCurveLoop()`, Revit gives a different set of objects, so the below `Where(x => x != externalLoop` query will give incorrect result:

https://github.com/BHoM/Revit_Toolkit/blob/4ff3d21861c83bf37a1d2c7aea05fb5cd18e95cd/Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs#L50

image

vietle-bh:

Closes #1499 

Instead of comparing objects, we can detect internal openings by their bounding boxes, which should be always smaller than the main external boundary's bounding box.

2024 loops fixed

### Test files

TBC

### DIFF
--- a/Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs
+++ b/Revit_Core_Engine/Convert/Geometry/FromRevit/Surface.cs
@@ -21,6 +21,7 @@
  */
 
 using Autodesk.Revit.DB;
+using BH.Engine.Geometry;
 using BH.oM.Base.Attributes;
 using BH.oM.Geometry;
 using System;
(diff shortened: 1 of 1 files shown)


vietle-bh:
@pawelbaran : Not anymore! I've changed to comparing bounding boxes instead. Please check again 🥂

pawelbaran (approved):
Reviewed and tested, the changes make sense, I really like the current approach 👍

pawelbaran:
@BHoMBot check required

vietle-bh:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

PR #1523, opened 2025-01-02
Add Revit 2025 support

### LINKED ISSUE #1520: Add support for Revit 2025
#### Description:

Revit 2025 uses .Net 8 instead of .Net Framework, and supporting it is necessary.

vietle-bh:

Closes #1520 

Added Revit 2025 support and tidied up csproj files in the process.

### Test files

Passing bot checks should be enough. We'll have the whole next few moths to test it 👍


.gitignore
```diff
@@ -51,6 +51,7 @@ bld/
 [Ll]og/
 [Bb]uild/
 [Pp]ackages/
+!libs/**/*
```
pawelbaran, line 54: I like it a lot! But why not simply `!libs/`?

Revit_Core_Engine/Convert/Structure/FromRevit/Profile.cs
```diff
@@ -862,7 +862,11 @@ private static BH.oM.Geometry.Line MullionLine(this Mullion mullion)
         /****              Private helpers              ****/
         /***************************************************/
 
-        private static readonly BuiltInParameterGroup[] dimensionGroups = { BuiltInParameterGroup.PG_STRUCTURAL_SECTION_GEOMETRY, BuiltInParameterGroup.PG_GEOMETRY };
+#if REVIT2020 || REVIT2021 || REVIT2022
```
pawelbaran, line 865: This is a bit risky because you seem to be changing the behaviour for Revit 2023 and 2024, am I correct? It may look a bit like an overkill, but the code below would allow us not to even think how does this part of the code work and whether the change would affect the functionality by any means
```suggestion
#if REVIT2020 || REVIT2021 || REVIT2022 || REVIT2023 || REVIT2024
```

Revit_Core_Engine/Create/Definition/Parameter.cs
```diff
@@ -49,9 +47,9 @@ public static partial class Create
         [Output("definition", "Revit parameter Definition created based on the input properties.")]
         public static Definition Parameter(Document document, string parameterName, string typeName, string groupName, bool instance, IEnumerable<string> categoryNames, bool shared, string discipline = "")
         {
-#if (REVIT2020 || REVIT2021 || REVIT2022)
+#if REVIT2020 || REVIT2021 || REVIT2022
             List<ParameterType> parameterTypes = new List<ParameterType>();
-            foreach (ParameterType pt in Enum.GetValues(typeof(ParameterType)))
+            foreach (ParameterType pt in System.Enum.GetValues(typeof(ParameterType)))
```
pawelbaran, line 52: why did you remove `using System` and add full qualification? that seems to be causing unnecessary noise...

Revit_Core_Engine/Create/Definition/Parameter.cs
```diff
@@ -118,6 +116,7 @@ public static Definition Parameter(Document document, string parameterName, stri
                 return Create.SharedParameter(document, parameterName, parameterType, groupName, instance, categories);
             else
             {
+#if REVIT2020 || REVIT2021 || REVIT2022
```
pawelbaran, line 119: Same as in my previous comment, I would consider including 2023 and 2024 here.

Revit_Core_Engine/Create/Definition/ProjectParameter.cs
```diff
@@ -42,10 +42,10 @@ public static partial class Create
         [Input("instance", "If true, the created parameter will be an instance parameter, otherwise it will be a type parameter.")]
         [Input("categories", "Categories, to which the created parameter is bound. It will get bound to all categories if this value is null.")]
         [Output("definition", "Revit project parameter Definition created based on the input properties.")]
-#if (REVIT2020 || REVIT2021 || REVIT2022)
+#if REVIT2020 || REVIT2021  || REVIT2022
         public static Definition ProjectParameter(Document document, string parameterName, ParameterType parameterType, BuiltInParameterGroup parameterGroup, bool instance, IEnumerable<Category> categories)
 #else
-        public static Definition ProjectParameter(Document document, string parameterName, ForgeTypeId parameterType, BuiltInParameterGroup parameterGroup, bool instance, IEnumerable<Category> categories)
```
pawelbaran, line 48: Same as previously, I would rather create another if/else for 2025 to avoid changing the original behaviour...

Revit_Core_Engine/Query/LookupParameter.cs
```diff
@@ -42,7 +42,11 @@ public static partial class Query
         [Input("parameterNames", "Names of the parameter to be iterated over in search for the parameter.")]
         [Input("parameterGroups", "If not null, only the Revit parameters from the given parameter groups will be parsed.")]
         [Output("parameter", "Parameter extracted from the input Revit element.")]
+#if REVIT2020 || REVIT2021 || REVIT2022
```
pawelbaran, line 45: Same as previously - in general, the comment concerns all places where you added an `#if` that spans only between 2020 and 2022 rather than 2024

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
```diff
@@ -0,0 +1,95 @@
...
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using System.ComponentModel;
+using BH.oM.Base.Attributes;
+#if !REVIT2020 && !REVIT2021 && !REVIT2022
```
pawelbaran, line 26: Could we make sure we do not invert the `#ifs`? Not sure if I am right in my concerns, but I am a bit afraid of us missing these by e.g. searching '#if revit2020' etc. In general, it feels a bit like an unnecessary complication to me.

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
```diff
@@ -0,0 +1,95 @@
...
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using System.ComponentModel;
+using BH.oM.Base.Attributes;
+#if !(REVIT2020 || REVIT2021 || REVIT2022)
```
pawelbaran, line 26: Why would these usings be harmful under, say, Revit 2023?

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            return def.ParameterGroup;
+        }
+#else
+        [Description("Gets the GroupTypeId of a Revit parameter definition.")]
+        [Input("def", "A Revit parameter definition to get GroupTypeId for.")]
+        [Output("groupTypeId", "GroupTypeId of a Revit parameter definition.")]
+        public static ForgeTypeId ParameterGroupTypeId(this Definition def)
+        {
+            return def.ParameterGroupTypeId();
```
pawelbaran, line 89: Isn't this method causing an infinite loop? Also if `ParameterGroupTypeId(string)` is meant to be called here, wouldn't it be better to just paste the body of that method here? This way you would not need `#if !(REVIT2020 || REVIT2021 || REVIT2022)` statement from what I see

pawelbaran:
> @pawelbaran : The below commit fixed Batch Build and allowed passing the related bot check. Please see if you support the changes. [924e81d](https://github.com/BHoM/Revit_Toolkit/commit/924e81d460e8a08bc8b8233b2d3ea7cf41ac0145)

The only dislike I have is this guy in VS 😃 But probably it can be hidden via UI.
image

On the side of code, we'll need to double check if the newly added attributes do not cause any issues downstream (`GenerateAssemblyInfo` gives me an irk in particular). But initial thumbs up!! 👍

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
pawelbaran: Using one of the favourite terms of @al-fisher, I have a feeling that tail starts to wag the dog here 😉 personally I am strongly against adding a compiler directive to fix an issue with VS setup - it causes unnecessary noise ('why is it here?') plus create a really poor precedent for the future behaviours.

Is there a chance you could tweak the IDE settings to avoid the issue? Alternatively you could fully qualify the types, but that won't look great either...

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
```diff
@@ -0,0 +1,95 @@
...
+        private static Dictionary<string, ForgeTypeId> m_GroupTypeIdByName = new Dictionary<string, ForgeTypeId>();
+
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Gets the GroupTypeId of a Revit parameter group by its name in the UI.")]
+        [Input("groupName", "Name in the UI of a Revit parameter group.")]
+        [Output("groupTypeId", "GroupTypeId of a Revit parameter group by its name in the UI.")]
+        public static ForgeTypeId ParameterGroupTypeId(this string groupName)
```
pawelbaran, line 49: I had a closer look at this method, I would consider migrating the code out of `Create.Parameter` as an overload for Revit 2020-24 returning `BuiltInParameterGroup`, then calling `ParameterGroupTypeId` for all versions of Revit inside `Create.Parameter`. Hopefully I did not confuse it too much 😉 Let's pick it up tomorrow on the catchup if unclear or not working 👍

pawelbaran: Yes, exactly!

pawelbaran:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

pawelbaran:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

pawelbaran:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

pawelbaran:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

Revit_Core_Adapter/Revit_Core_Adapter.csproj
```diff
@@ -1,248 +1,235 @@
...
+		</Reference>
+		<Reference Include="UI_oM">
+			<HintPath>$(ProgramData)\BHoM\Assemblies\UI_oM.dll</HintPath>
+			<Private>False</Private>
+			<SpecificVersion>False</SpecificVersion>
+		</Reference>
+	</ItemGroup>
+
+	<ItemGroup>
+		<None Include="Listener\Files\BHoM_2020.Addin">
```
pawelbaran, line 188: This could be parametrised, also 2025 is missing

Revit_Core_Engine/Create/Definition/Parameter.cs
```diff
@@ -22,11 +22,12 @@
...
 using Autodesk.Revit.DB;
 using BH.oM.Base.Attributes;
-using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
-using System.Reflection;
+#if REVIT2020 || REVIT2021 || REVIT2022 || REVIT2023 || REVIT2024
+using System;
+#endif
```
pawelbaran, line 30: Please stop wrapping usings in #ifs

Revit_Core_Engine/Query/LookupParameter.cs
```diff
@@ -51,9 +55,14 @@ public static Parameter LookupParameter(this Element element, IEnumerable<string
             {
                 foreach (Parameter p in element.Parameters)
                 {
+#if REVIT2020 || REVIT2021 || REVIT2022 || REVIT2023 || REVIT2024
+                    BuiltInParameterGroup groupType = p.Definition.ParameterGroup;
+#else
+                    ForgeTypeId groupType = p.Definition.GetGroupTypeId();
+#endif
```
pawelbaran, line 62: Couldn't this become something around `var groupType = p.Definition.GroupTypeId()`?

Revit_Core_Engine/Query/ParameterGroupTypeId.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Query
+    {
+#if REVIT2020 || REVIT2021 || REVIT2022 || REVIT2023 || REVIT2024
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        public static BuiltInParameterGroup ParameterGroupTypeId(this string groupName)
```
pawelbaran, line 40: Looking at similar cases of `Spec` and `Unit`, I would argue there should be 3 methods:
- `GetGroupTypeId(Definition)` returning the enum or `ForgeTypeId`
- `GroupName` returning name of  the property of `GroupTypeId` static class that is correspondent to the provided value
- `GroupFromName` - inverse of the above

...please see similar methods for `Unit` for reference. If we need label on top of the above, let's add another method (the one you added seems to be OK), but any string-based matching should happen around the property names of the static class that lists out `ForgeTypeIds` of given category.

pawelbaran (changes requested):

Revit_Core_Engine/Query/LookupParameter.cs
pawelbaran: Yes that's true, personally I treated such cases as 'different' - most likely wrongly, should be avoiding exceptions. Then maybe an alternative, `dynamic` instead of `var`?

Revit_Core_Engine/Query/GroupTypeId.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Output("parameterGroup", "BuiltInParameterGroup of a Revit parameter definition.")]
+        public static BuiltInParameterGroup GroupTypeId(this Definition def)
+        {
+            return def.ParameterGroup;
+        }
+#else
+        [Description("Gets the GroupTypeId of a Revit parameter definition.")]
+        [Input("def", "A Revit parameter definition to get GroupTypeId for.")]
+        [Output("groupTypeId", "GroupTypeId of a Revit parameter definition.")]
+        public static ForgeTypeId GroupTypeId(this Definition def)
```
pawelbaran, line 47: I think this method could be removed with version for 2020-24 to be renamed to `GetGroupTypeId`, same as done with `GetDataType` or `GetUnitTypeId`. Will approve this PR and raise an issue to be picked up in a separate thread.

pawelbaran (approved):
Approving with the caveat as in the comment

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check ready-to-merge

PR #1526, opened 2025-01-14
Rename GroupTypeId Method

### LINKED ISSUE #1525: Rename GroupTypeId to GetGroupTypeId and leave only for versions that do not have native API method
#### Description:

Please see [this comment](https://github.com/BHoM/Revit_Toolkit/pull/1523#discussion_r1915187150)

vietle-bh:

Closes #1525

Renamed GroupTypeId Method to GetGroupTypeId

### Test files

Installer: 
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/%231526-RenameGroupTypeIdMethod?csf=1&web=1&e=aoJ5Xo

### DIFF
--- a/Revit_Core_Engine/Query/GetGroupTypeId.cs
+++ b/Revit_Core_Engine/Query/GetGroupTypeId.cs
@@ -32,7 +32,7 @@ namespace BH.Revit.Engine.Core
 {
     public static partial class Query
     {
-#if REVIT2021 || REVIT2022 || REVIT2023 || REVIT2024
+#if REVIT2021
         [Description("Gets the BuiltInParameterGroup of a Revit parameter definition.")]
         [Input("def", "A Revit parameter definition to get BuiltInParameterGroup for.")]
         [Output("parameterGroup", "BuiltInParameterGroup of a Revit parameter definition.")]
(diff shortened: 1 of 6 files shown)


Revit_Core_Engine/Query/GetGroupTypeId.cs
```diff
@@ -36,15 +36,15 @@ public static partial class Query
...
+        public static BuiltInParameterGroup GetGroupTypeId(this Definition def)
         {
             return def.ParameterGroup;
         }
 #else
         [Description("Gets the GroupTypeId of a Revit parameter definition.")]
         [Input("def", "A Revit parameter definition to get GroupTypeId for.")]
         [Output("groupTypeId", "GroupTypeId of a Revit parameter definition.")]
-        public static ForgeTypeId GroupTypeId(this Definition def)
+        public static ForgeTypeId GetGroupTypeId(this Definition def)
```
pawelbaran, line 47: I think this method could be removed altogether, so that you can simply call `Definition.GetGroupTypeId()` everywhere, with usings being `BH.Revit.Engine.Core` for Revit 2024 and lower and `Autodesk.Revit.DB` for 2025 and above

pawelbaran:
Hmm thinking... this method is new, so should be OK to change the compiler directive to only compile this method only against versions where it does not exist in Revit API?

pawelbaran:
After the discussion with @vietle-bh, I am afraid I was wrong about the requested change from the very inception:
- in Revit 2021-24 all code is based in `BuiltInParameterGroup` enum
- `GetGroupTypeId` exists in Revit API from version 2022 onwards and it returns `ForgeTypeId` across all versions
- that means that if we rename `GroupTypeId` method to `GetGroupTypeId` and try to Revit API, we will suffer from one of the two:
    a) conflict between `BuiltInParameterGroup` and `ForgeTypeId` if we implement it only for 2021
    b) ambiguity between native Revit API method and our extension method if we implement it for versions 2021-24

Apologies for the confusion, as agreed with @vietle-bh, we will stick to what we currently have on `develop`, so the PR is going to be closed.

PR #1532, opened 2025-02-07
Add BHoM Filters for optimisation on Excel Add-in

linhnam-nguyen:
### NOTE: Depends on 
Require PR: Excel UI -> https://github.com/BHoM/Excel_UI/pull/406
New dependencies to Revit_Engine, Nuget Packages:
- Microsoft.CodeAnalysis
- Microsoft.CodeAnalysis.CSharp

   
### Objectif
To create a comprehensive set of tools for BIM managers, enabling:

- Free exploration of models with highly flexible data manipulation capabilities.
- The creation of simple, communication-friendly spreadsheets for project managers.

This toolkit is designed for BIM managers with some coding and Excel skills, reducing reliance on third-party software. It allows for faster development of new in-house toolboxes without the need for extensive software development expertise. Examples of potential application:

- Externalize geometry constraints into an Excel file, centralizing family datasheets and lookup tables in a single location. This facilitates standardization and prevents the creation of unnecessary parameters within families, leading to cleaner, more efficient Revit files with a smaller footprint.
- Establish constraints among parameters of System Family elements like ducts and pipes, enabling custom calculations and verification methods tailored to specific needs.
- Automate classification processes, perform lookups, and assign elements based on conditions, significantly streamlining the takeoff process or documentation and improving overall productivity.
- Out-source and manage information of equipment's data to a spreadsheet, promote use of generic objects for early stage of design, this will facilitate project manager or engineers to focus on their research and design, rather than manipulation or tracking metadata model.

### Test files

Excel and Revit (french version) test files: 
[Test File.zip](https://github.com/user-attachments/files/18765394/Test.File.zip)

### Changelog

Feature: ParameterFormula
- Add `IsInstanceParameter` method to check on a RevitParameter of a BHoM object,
- Add `LabelForSymbolTypeId` method to show display unit of RevitParameter in BHoM object
- Add new properties to `RevitParameter`: IsInstance, DisplayUnit
- Create a new BHoM object `ParameterFormula` to hold relationships among the Revit Parameter
- Add `CreateFormula` methods to compile in runtime and return a delegate from a built-in Templates
- Add `ApplyFormula` methods to calculate ParameterFormula and transfer results to RevitParameter objects
- Add `GetRevitParameter` method to retrieve a RevitParameter object of a BHoM entity.

Features: New methods to RevitEngine project to improve Excel UX:
Reason for this proposal: Reduce intensive pull requests operation needs for Excel user therefore more responsive work on large models
- Add `GetRevitParameterValues `to investigate on a List of BHoM objects to reduce compute need and for a better organized spreadsheet.
- Add `FilterUniqueRevitElements` : In case of MEPCurve, an object might be segmented into small segments while RevitParameter values unchanged and causing incorrect takeoff or aggregation.
- Add `FilterByRevitParameter` and `FilterByRevitParameters` that can handle multiple types of filters, based on IRevitParameterFilter.
- Add `SortByRevitParameter` to organize pulled objects.

### Issues may happen:

These developments are working on Windows10 and Windows 11 machines with Revit version 2023:
 `CreateFormula` may not work because of cached assemblies of compiler, worked solution for me:
 https://learn.microsoft.com/en-us/answers/questions/1302681/could-not-load-file-or-assembly-system-runtime-com 

### Example of spreadsheet design:
A Settings Sheet to control all requests:
image
We can configure a spreadsheet to pull all necessary objects at once for a working session:
image
GetRevitParameterValues makes it able to use one formula cell per column. By using Excel array syntax for dynamic array references, the spreadsheet is optimized for startup time and calculation.
image
Pivot table is a useful feature to take a snapshot of pulled data. **Data is only renewed on demand. **
image
image


linhnam-nguyen:
@BHoMBot check compliance

Revit_oM/Parameters/RevitParameter.cs
```diff
@@ -43,9 +43,15 @@ public class RevitParameter : IObject
         [Description("Unit type of the Revit parameter.")]
         public virtual string UnitType { get; set; }
 
+        [Description("Current Using Unit of the Revit parameter.")]
+        public virtual string DisplayUnit { get; set; }
+
         [Description("Whether the parameter is read only or modifiable by the Revit user.")]
         public virtual bool IsReadOnly { get; set; } = false;
 
+        [Description("Whether the parameter is instance parameter.")]
```
pawelbaran, line 52: This should not be needed, type parameters are already fetched and stored on the property correspondent to element type, e.g. for class `Wall` it is `Construction`:
image

Revit_Engine/Query/GetRevitParameterValues.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Retrieves values of a parameter attached to a groupe of BHoM objects. It will return null values for objects that do not have  the parameter defined")]
+        [Input("List<bHoMObject>", "BHoMObjects to which the parameters will be attached.")]
+        [Input("parameterName", "Name of the parameter to be sought for.")]
+        [Output("List of values in same order as objects' list")]
+        public static List<object> GetRevitParameterValues(this List<IBHoMObject> bHoMObjects, string parameterName)
```
pawelbaran, line 42: Is that needed to improve the UX in Excel? There already is a similar method `GetRevitParameterValue`, which yields exactly same results in GH, as far as I understand

Revit_Engine/Query/IFilterServices.cs
```diff
@@ -0,0 +1,266 @@
...
+                foreach (var crit in criteria)
+                {
+                    IRevitParameterFilter filter = CreateFilter(crit.FilterType);
+                    bHoMObjects = bHoMObjects.Where(x => filter.IsMatch(x, crit.ParameterName, crit.ParameterValue));
+                }
+                return bHoMObjects.ToList();
+            }
+        }
+
+        public interface IRevitParameterFilterService
```
pawelbaran, line 50: It feels like it could all be replaced with a bit of play with the framework around `ICondition`

pawelbaran (commented):
Hi @linhnam-nguyen,

First of all, thank you for the great contribution and apologies for being silent for quite a few days, but I was literally working on something that may help you a lot 🙈 please have a read below.

I am not sure if I capture the big picture and your ultimate objective, so please correct me if I am wrong. What I think you are trying to achieve is to build a mechanism that would allow you to work with the pulled objects by updating their parameter values based on certain formulas that would be defined by the user.

If I am correct about the above, I am afraid that we missed each other literally by a few weeks! Recently I was working on two things that may be particularly helpful in your workflows:
- Verification framework, which introduces the concept of `ICondition` (what you most likely are looking for is [`ValueCondition`](https://github.com/BHoM/BHoM/blob/develop/Verification_oM/Conditions/ValueCondition.cs)), `IValueSource` and execution of human-created formulae (see [this method](https://github.com/BHoM/BHoM_Engine/blob/88d79af29451120bf14feb2263f16ea4a142455e/Verification_Engine/Compute/VerifyCondition.cs#L404))
- extension of the above general framework for Revit merged literally this week: https://github.com/BHoM/Revit_Toolkit/pull/1533, which refactors parameter filtering to leverage Verification (comparison type enum, concept of `ICondition` as well as `IValueSource`, see [ParameterValueSource](https://github.com/BHoM/Revit_Toolkit/blob/develop/Revit_oM/Parameters/ParameterValueSource.cs)

The above should give you most of the functionality you need I believe:
- an abstract concept of extracting parameter values via `ParameterValueSource`
- filtering by parameter value based on `ValueCondition` combined with `ParameterValueSource` and `Passes` method
- cornerstone for formula evaluation mechanism - here you'll probably need to extend it a bit (check if what I did with tolerance in `Solve` makes sense in your use case, add a `Formula` object, make that method public etc.)

Following the above, I think you should not need to add much to Revit_Toolkit (`ParameterValueSource` can be used to both read and write parameter values really - for the latter you'd just need to add `TrySetValue` similar to `TryGetValueFromSource`). Where the code surely needs a bit of massaging is formula execution as well as a few smart methods that would enable the exact workflow you are trying to achieve.

I am wondering about how to continue this thread, I would like to make it as efficient for you as possible - first of all, could you please reflect on the above as well as the code comments I left? If it all feels to confusing, maybe we could try to grab a call to streamline the process.

Anyways, good stuff, really excited to make it to a happy end! 👍

Revit_oM/Parameters/RevitParameter.cs
```diff
@@ -43,9 +43,15 @@ public class RevitParameter : IObject
         [Description("Unit type of the Revit parameter.")]
         public virtual string UnitType { get; set; }
 
+        [Description("Current Using Unit of the Revit parameter.")]
```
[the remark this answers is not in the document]
linhnam-nguyen: The screenshots below illustrate how I use this property:
DisplayUnit helps users by showing the unit symbol of the value they are working with.
image
UnitType represents the domain (discipline) of the parameter, but I’m not entirely sure about its practical use case. Could you clarify? I’m genuinely curious to understand its purpose.
image

When users work with objects, they might be confused by the unit system used in BHoM. In HVAC, for example, duct airflows are commonly measured in m³/h, and pipe flow calculations often use L/s or m³/h, whereas BHoM uses m³/s. Having access to the current unit ensures that users apply the correct conversion ratio in their formulas, preventing errors and improving clarity.

Revit_Engine/Query/GetRevitParameterValues.cs
linhnam-nguyen: While I don't work with Grasshopper, I assume it functions similarly to Dynamo, where if a node can operate on a single type, it can also work on a collection of that type.

As I’ve been testing this, I’ve come to realize that methods rendered from BHoM don’t support dynamic input types in ExcelUI. Without explicitly creating a method that can handle collections, we would need to define a function for each element in a collection, which would lead to an unscalable spreadsheet.

To elaborate:

If Cell A1 contains a List<Object>, calling:
`Query.Adapters.Revit.GetRevitParameterValue?by_IBHoMObject_String(A1, parameterName)` is not allowed

Instead, we would have to:

1. Expand the list from A1 into multiple cells (e.g., starting from A2).
2. Apply the function iteratively for each element:

`Query.Adapters.Revit.GetRevitParameterValue?by_IBHoMObject_String(A2, parameterName)`
`Query.Adapters.Revit.GetRevitParameterValue?by_IBHoMObject_String(A3, parameterName)`
...
Until the end of the list.

From my understanding, this approach results in high computational intensity due to the execution time of the `ToExcel ` method and multiple relection calls of `GetRevitParameterValue`. Additionally, on startup, ExcelUI scans all used cells in the workbook to perform calculations, meaning that if a spreadsheet contains a lot of cells using BHoM functions, the user would experience significant delays.

To solve this, I created a method that directly works with collections, allowing us to simply use:
`Query.Adapters.Revit.GetRevitParameterValues?by_ListOfIBHoMObject_String(A1, parameterName)`

Alternatively, with Excel Array syntax (available in Office 365), if the list is expanded in A2, we can use:
`Query.Adapters.Revit.GetRevitParameterValues?by_ListOfIBHoMObject_String(A2#, parameterName)`
This makes the spreadsheet more optimized.

I’d love to hear your thoughts on this approach

Revit_Engine/Query/IFilterServices.cs
linhnam-nguyen: Though I know this approach doesn’t fully align with the codebase style, I don’t have a complete view of all BHoM base repo. For now, I’ve temporarily placed this service in Revit_Engine to expose it directly to the UI while limiting its impact.

Thanks for the `ICondition` reference—I'll take a closer look and try to figure it out.

linhnam-nguyen: Hello @pawelbaran, thank you for your reply. While working on this feature locally, I realized it would take time to properly structure everything in the codebase. This became even more apparent when I saw the recent updates on the develop branch, including recent snippet codes preparing for this use case. I anticipate a lot of work ahead, but I’m really excited about it!

For this PR, my goal is to provide users with a flexible way to implement their own data logic or constraints within the model using user-defined methods—whether through simple mathematical formulas or multi-step calculations. My initial focus is on classification processes for takeoff and documentation.

Since I don’t have a complete view of all BHoM projects, I initially chose to limit this feature to the Revit Toolkit, even though it could be generalized for many other use cases in the future. With your recent developments, I’m glad to see that this feature can now have broader applications. I’ll take some time to review the latest updates and will return in a few weeks, hopefully with a more refined implementation.

Regarding this PR, I believe it would be best to split it into two independent PRs:
- This PR will focus solely on small improvements based on Excel experience and my limited insight, such as data preprocessing (sorting, filtering, etc.).
- A separate PR will be opened later for the core implementation, once I have studied your code in more depth.

linhnam-nguyen:
Hello @pawelbaran,

Thank you for your reply! While working on this feature locally, I realized that properly structuring everything in the codebase would take time. This became even more evident when I saw the recent updates on the develop branch, including new snippet codes preparing for this use case. There’s a lot of work ahead, but I’m really excited about it!

For this PR, my goal is to provide users with a flexible way to implement their own data logic or constraints within the model using user-defined methods—whether through simple mathematical formulas or multi-step calculations. My initial focus is on classification processes for takeoff and documentation.

Since I don’t have a complete view of all BHoM projects, I initially chose to limit this feature to the Revit Toolkit, even though it could be generalized for many other use cases in the future. With your recent developments, I’m glad to see that this feature can now have broader applications. I’ll take some time to review the latest updates and will return in a few weeks, hopefully with a more refined implementation.

Regarding this PR, I believe it would be best to split it into two independent PRs:

- This PR will focus solely on small improvements based on Excel experience and my limited insight, such as data preprocessing (sorting, filtering, etc.).
- A separate PR will be opened later for the core implementation, once I have studied your code in more depth.

Please let me know your thoughts.

Revit_oM/Parameters/RevitParameter.cs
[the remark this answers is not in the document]
pawelbaran: The reason for using quantities rather than units is to keep the BHoM as convention-agnostic as possible, i.e. all values are stored in base SI units (with one quirk, kg 🤯), units being dealt with on the UI level. In the context of a practical example, if we now add `DisplayUnits` property to `RevitParameter` and pull a 1 meter long wall from Revit, where length in the Revit UI is represented in mm, the user will have the following:
- `Value = 1`
- `UnitType = Length`
- `DisplayUnits = mm`

If I saw data in such structure without context, I would expect the wall to be 1mm long.

However, I see the practical reasons for adding the property, so would not be against if that would become a blocker to your workflow 👍

Revit_Engine/Query/GetRevitParameterValues.cs
pawelbaran: Ah I see! I am not an expert in Excel UI, but now started thinking if that's not something that should be solved on the level of UI rather than in the code... @adecler do you know if it would be hard to enable Excel execute methods taking individual objects against unexploded collections stored in a single cell?

Revit_Engine/Query/IFilterServices.cs
pawelbaran: Yeah it is perfectly fine to put it this way while prototyping - this is exactly why we have the review process, we always can tweak the code before merging 👍 But still, I think you will not need this interface at all if you leverage the recently added stuff

linhnam-nguyen:
Hi @pawelbaran, I've just come across your recently merged commits in Revit Toolkit and BHoM Engine.

I somewhat understand your intention to unify filters in requests and queries by using ICondition for filtering parameter values in Revit Toolkit. However, regarding `FilterByParameterExistence`, is this a work in progress with a new approach, or is it being removed from the Toolkit?

In my current workflow, I use this feature to distinguish between instance and type elements based on the `Phase Created` parameter. If there’s another way to achieve this, I’d love to learn about it. Otherwise, if it makes sense, I’m considering reimplementing it using `ElementIdsByCondition`—though I’m unsure if that’s the best approach, since FilterByParameterExistence isn’t a `ConditionRequest`.

Regarding `FormulaCondition`, I really like the lean approach you’ve taken—it makes formulas much more readable and intuitive. However, in my case, I see a few challenges unless the Solve method is modified. Please correct me if I’m mistaken—I’m still learning, but here’s what I’ve noticed:

- The syntax supports basic mathematical operators and `if/else` logic, which is great.
- Methods from `System.Math`, string operations, and LINQ expressions don’t seem to be supported at the moment.
- When testing runtime code execution using CSharpScript in my prototype, I ran into memory leak issues due to the lack of GC cleanup for assemblies ( https://github.com/dotnet/roslyn/issues/41722 ). I worry this could become unsustainable for large models with heavy calculations.
- Performance concern: Since formulas are re-evaluated on each iteration, there might be performance issues if there’s no caching mechanism in place, so that I use Roslyn Emit method on memory with a cache mechanism.

I’d love to hear your thoughts on this and see if there are ways to improve flexibility and optimize performance. I also sent you a LinkedIn request in case that’s easier to discuss in details. Let me know what works for you.

Looking forward to your insights.

linhnam-nguyen:
Hi @pawelbaran, I´m back. Thank you for your response.


> This object is still in the oM, not to be deleted in this milestone, but I am on the fence whether not to replace it with a `ConditionRequest` holding `BH.oM.Verification.Conditions.HasValue`. This, however, would return same results for both empty parameter and no parameter, therefore I am still not sure if this is the right approach - happy to hear your opinion, could help in making the final decision 👍
> 

That´s great to hear.

> There definitely are other ways to achieve this, for example adding `IsElementType(this Element element)` extension method, which then could be called when processing `ValueCondition` with value source being `PropertyValueSource`:
> 
> ```cs
> ValueCondition condition = new ValueCondition
> {
>     ValueSource = new PropertyValueSource
>     {
>         PropertyName = "IsElementType"
>     },
>     ReferenceValue = false,
>     ComparisonType = ValueComparisonType.EqualTo
> }
> ```

 I’ll try it right away and see how it works in my case.


> Wow that is interesting! I have not heard of that issue and never faced it when doing my own testing, nor writing a simple app on top of the formula mechanism. Is there a chance you could share a console app that would highlight the issue? Naturally it should be fixed.
> 

The issue with memory leaks in CSharpScript has been consistent in my tests. You can try running this simple console app to analyze the issue:
 ```cs
using System.Reflection;
using Microsoft.CodeAnalysis;
using ConsoleApp1;
using Microsoft.CodeAnalysis.CSharp.Scripting;
using Microsoft.CodeAnalysis.Scripting;


public static partial class MethodGenerator
{
    public class FormulaGlobals
    {
        public Dictionary<string, object> Variables { get; set; } = new();

        public T Get<T>(string key)
        {
            if (Variables.TryGetValue(key, out var value))
            {
                return (T)Convert.ChangeType(value, typeof(T));
            }
            throw new KeyNotFoundException($"Variable '{key}' not found.");
        }
    }

    public static async Task<object> FormulaScripting(string codeString, Dictionary<string, object> variables = null)
    {
        var globals = new FormulaGlobals { Variables = variables ?? new Dictionary<string, object>() };

        var options = ScriptOptions.Default
            .AddReferences(
                typeof(object).Assembly,
                typeof(Console).Assembly,
                typeof(Dictionary<,>).Assembly,
                typeof(Enumerable).Assembly
            )
            .AddImports("System", "System.Linq", "System.Collections.Generic");

        return await CSharpScript.EvaluateAsync(codeString, options, globals, typeof(FormulaGlobals));
    }
}

class Program
{
    static async Task Main()
    {
        long memoryBefore = GC.GetTotalMemory(true);
        Console.WriteLine($"Memory Before: {memoryBefore / 1024} KB");

        for (int i = 0; i < 1000; i++)
        {
            string formula = "Get<int>(\"x\") + Get<int>(\"y\")";
            var variables = new Dictionary<string, object>
        {
            { "x", 10 },
            { "y", 20 }
        };
            object result = await MethodGenerator.FormulaScripting(formula, variables);
            GC.Collect(); // Force garbage collection
            long memoryAfter = GC.GetTotalMemory(true);
            Console.WriteLine($"Memory After {i} iterations: {memoryAfter / 1024} KB");
        }
        long memoryAfterGC = GC.GetTotalMemory(true);

        Console.WriteLine($"Memory After GC: {memoryAfterGC / 1024} KB");

    }
}
 ```
image
image

As you can see, I can not use GC collector on CScript. The memory usage keeps increasing. 

> > * Performance concern: Since formulas are re-evaluated on each iteration, there might be performance issues if there’s no caching mechanism in place, so that I use Roslyn Emit method on memory with a cache mechanism.
> 
> Agreed, performance is far from great atm, is there a chance you could shed a bit more light on how are you using Roslyn Emit? As above, defo something to improve if possible 👍

Regarding Roslyn emit, you can check out in my commit : Revit_Engine/Compute/CreateFormula.cs.

To summarize my approach:

1. Formula Preparation: I inserted the user’s formula string into a predefined template, creating a static method inside a `CustomMethod `class. I included commonly used namespaces (such as System.Linq) to allow users to access built-in methods.
2. Dynamic Input Handling: I stored variables as key-value pairs (Dictionary<string, object>) where users define both variable names and types. These were then converted into method parameters inside the generated code. This was inspired by your CSharpToolkit but implemented in a more direct and simplified way.
3. Memory Compilation: The generated code is compiled in-memory and returned as a delegate for execution. This ensures that GC (Garbage Collector) can properly clean up resources, preventing memory leaks.
4. Caching Mechanism: If compilation succeeded, the entire code string was stored in a cache dictionary (Dictionary<string, Delegate>) to avoid redundant compilation in future executions.

This approach gives power users complete flexibility, allowing them to define single-line or multi-line calculations. However, the downside is that debugging can be difficult, especially since BHoM lacks a robust logging mechanism. When testing in BHoM Excel, I also noticed that error messages are limited to 255 characters, which means longer debugging messages get truncated. I had to trim error logs to only essential details, which made troubleshooting more difficult.

> But first, would be great if we could narrow down the problem to formula evaluation - please give it a shot to build as much as you can using the objects and methods already available in BHoM, so that we could only focus on the gaps rather than the entire architecture at the same time. Cheers!

Totally agree. I'm exploring ways to enhance your evaluation framework and refine the approach. Hopefully, we can find a successful path forward for this feature.

linhnam-nguyen:
Hi @pawelbaran,

I've recently updated the PR with the modifications we discussed. Here’s what changed in this update:

- Removed the `IsInstance `property from `RevitParameter`,
- Removed `ParameterFormula `(to be included in a future PR) ,
- Adapted `FilterByRevitParameters `to work with `ICondition `objects and Verification mechanism.

I’m looking forward to your thoughts. I think this PR is now in a good state to be reopened for review.
Thanks!

pawelbaran (commented):
Thanks @linhnam-nguyen for another push 👍 

This time I flipped the roles a bit and [branched off your PR](https://github.com/BHoM/Revit_Toolkit/tree/Revit_Toolkit-%231532-Review) to try and restructure your code how I would do it. The code is not perfectly rounded but should show my intent well enough - it would be great if you could have a look and tell if this direction would work for you or I went too far with my cleanup. Main changes I made:
- renamed the properties in `RevitParameter` class to make them a bit easier to read for he users
- removed `out` parameter from `ParameterValue` method, also reshuffled the label query methods
- simplified `TryGetValueFromSource` method
- simplified `FilterByRevitParameter` method and made it only accept `ValueComparisonType` enums as I do not think it is a good idea to do custom string parsing inside Revit_Toolkit - as a core repo it should stay clean of any customisation, so that part should be a part of the Excel workflow, and if that is not feasible then I would argue we need a new Engine project that would do this sort of juggling
- replaced `FilterUniqueRevitElements` with a more generic `Unique` method

What is still hanging and requires another iteration:
- `GetRevitParameter` vs `GetRevitParameters` vs `GetRevitParameterValue` vs `GetRevitParameterValues` seems to be a bit inconsistent in terms of returned results - some methods return property values and the others not
- as commented before, I think `GetRevitParameterValues` should not be needed, it should be solved on the side of Excel_UI by enabling function calls against each item of the input collection (will chase it internally to check how feasible is that) - in the meantime, I would recommend you creating your personal repo that would contain an Engine project where you could dump all such utilities as well as prototype your workflows faster.

Please let me know how does it all sound - if you like the direction I took on my branch, I am happy to polish the code and raise a PR for you to review, meanwhile you could open a personal repo with utility methods to keep the momentum going.

pawelbaran:
@linhnam-nguyen please also have a look at https://github.com/BHoM/Excel_UI/issues/333

linhnam-nguyen:
Hi @pawelbaran , and thank you for taking the time to go through my code.The improvements you made make a lot of sense, and I really appreciate the clarity you've brought in. Below are some of my remarks and clarifications regarding your modifications:

1. Separation of `DisplayUnit `(or `UnitLabel)` from `ParameterValue` Your approach is much cleaner and more concise than mine. When I originally implemented it, I was trying to minimize the impact on the existing codebase and reduce direct access to the Revit API to help with performance.

2. `FilterByCondition `/ `TryGetValueFromSource `Your simplification is elegant, but it shifts more responsibility to the modeler. My initial goal was to make filtering more forgiving, allowing users to filter on either type or instance parameters, depending on what's available.

I was thinking of a scenario where users (or me) have many mechanical components from different manufacturers (with key info stored in type parameters). They may not know whether the data is stored on the type or instance level—they just want the results.

To support that flexibility, what do you think of introducing a third option like "any" or "both" for valueSource.FromType, using a nullable boolean? That way:

- true = type only
- false = instance only
- null = allow both

Would love to hear your take on that!

3. `FilterUniqueRevitElements ` (`Unique`) I haven’t fully understood how this would now work. Could you show me an example of how to use it via the Grasshopper UI?

This function was originally written for MEP BHoM objects, where elements like pipes or ducts are segmented at each change in direction or cross-section—but the underlying Revit element remains the same. That creates confusion for users trying to synthesize quantities, so I needed a way to collapse those into unique Revit references.

4. `GetRevitParameterValues `You’re right—this ties into a larger, open issue in the Excel_UI repo. It’s a deeper subject if we want the logic to scale in a truly intuitive and dynamic way.

That said, I found a workaround using Excel’s native lambda + array functions:
`=MAP(A1:A4, LAMBDA(x, GetRevitParameterValue(x)))`

This helps bypass the current limitations while still letting me apply the function across a range.

Since you’ve opened a fresh branch, I’d be happy to use this as a prototype and continue our discussion while you continue refining the review PR.

Thanks again! Looking forward to your thoughts.

pawelbaran:
@linhnam-nguyen all sounds good! I agree without any comments with everything not listed below 😉 

> 2. `FilterByCondition `/ `TryGetValueFromSource `Your simplification is elegant, but it shifts more responsibility to the modeler. My initial goal was to make filtering more forgiving, allowing users to filter on either type or instance parameters, depending on what's available.
> 
> I was thinking of a scenario where users (or me) have many mechanical components from different manufacturers (with key info stored in type parameters). They may not know whether the data is stored on the type or instance level—they just want the results.
> 
> To support that flexibility, what do you think of introducing a third option like "any" or "both" for valueSource.FromType, using a nullable boolean? That way:
> 
> * true = type only
> * false = instance only
> * null = allow both
> 
> Would love to hear your take on that!

This one is a bit tricky, personally I feel a bit against introducing this sort of ambiguity (always harder to manage/maintain), but I ran it against @michal-pekacki who agreed with your request, I am becoming outvoted 😢 However, assuming that it is you who sets up the spreadsheets for the rest of the office, couldn't you wrap the condition into two identical `LogicalOrConditions` where one takes param value from instance and the other from type? I guess as long as it works for simple conditions, it may not be perfect for the larger ones?

> 3. `FilterUniqueRevitElements ` (`Unique`) I haven’t fully understood how this would now work. Could you show me an example of how to use it via the Grasshopper UI?

Ahh sorry, I thought you want to cull out the exact duplicates from your collection - now I see what I proposed does not help. Let me think how to do it in an elegant/scalable way.

> Since you’ve opened a fresh branch, I’d be happy to use this as a prototype and continue our discussion while you continue refining the review PR.

Feels like we're heading to an agreement 😄 I will try to fully convince myself about 2. and think about 3. - once that happens, I will finalise the work I did on my branch. Please tell me one thing: how urgent is this? If very, I will try to have a look later this week, otherwise it will easily go into the next one...

PR #1534, opened 2025-03-13
Naming of pulled physical/arch elements unified

### LINKED ISSUE #1519: Unexpected/unwanted type update in case of 2 types with same name
#### Description:

When updating elements via Revit adapter, type can be changed unexpectedly in case of multiple families containing types with name matching the type name of the updated element. That happens because on push/update Revit adapter attempts to update the type, and the default type picking mechanism is based on name of the property (the object returned by `GetRevitElementType` method), where the adapter picks first type that matches it.

I scratched my head a bit around that one, looks like the best solution will be adding family name prefix to the type name in name of the property on pull, i.e. after pulling, object returned by `GetRevitElementType` should have `Name` in the following format: `{familyName}: {typeName}` rather than `{typeName}` - that would eliminate the ambiguity.

#### Test files:
On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FRevit%5FToolkit%2F%231519%2DUnexpectedTypeUpdate)

pawelbaran:

Closes #1519

### Test files

See #1519.

### Additional comments

Originally I expected the problem to occur to most types in Revit_Toolkit, but apparently columns and framing were almost the only pushable types that would suffer from it - so luckily the change is quite surgical and safe, as far as I can think of potential consequences.

### DIFF
--- a/Revit_Core_Engine/Convert/Architecture/FromRevit/Opening.cs
+++ b/Revit_Core_Engine/Convert/Architecture/FromRevit/Opening.cs
@@ -57,7 +57,7 @@ namespace BH.Revit.Engine.Core
             if (opening != null)
                 return opening;
             
-            opening = new oM.Architecture.BuildersWork.Opening { Name = instance.Name };
+            opening = new oM.Architecture.BuildersWork.Opening { Name = instance.FamilyTypeFullName() };
 
             //Set coordinate system
             opening.CoordinateSystem = instance.CoordinateSystem();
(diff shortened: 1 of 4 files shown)


pawelbaran:
Thanks @vietle-bh, and what if you set `Push` to `true` in the 1st snip? In both cases `success` should be `true` - it indicates that push occurred without issues, not necessarily leading to type update (can also be location update, param etc.)

pawelbaran:
@BHoMBot check ready-to-merge

PR #1543, opened 2025-03-20
Filtering by ElementId fixed

### LINKED ISSUE #1542: Filtering by ElementId does not work
#### Description:

After the recent refactor of parameter filtering mechanism, it turned out that filtering by `ElementId` stopped working. That is because inside parameter value comparison, `ParameterValue` query is called via `TryGetValueFromSource`, resulting in returning a string representation of an element under given parameter rather than the integer value of id itself. Then that string representation is compared with integer id coming from the request, obviously never matching.

https://github.com/BHoM/Revit_Toolkit/blob/49d3671c38364f86666ff5c4cd352e5f3dc07640/Revit_Core_Engine/Compute/TryGetValueFromSource.cs#L83-L103

The mechanism needs to be tweaked to return integer value of element id instead of string.

pawelbaran:

Closes #1542

### Test files

Standard requests => filtering by elementId and parameters. Plus UT on top of that.

### DIFF
--- a/Revit_Core_Engine/Compute/TryGetValueFromSource.cs
+++ b/Revit_Core_Engine/Compute/TryGetValueFromSource.cs
@@ -75,33 +75,6 @@ namespace BH.Revit.Engine.Core
             return new Output<bool, object> { Item1 = true, Item2 = param.ParameterValue() };
         }
 
-
-        /***************************************************/
-        /****              Private Methods              ****/
-        /***************************************************/
-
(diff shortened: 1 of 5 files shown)


pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check ready-to-merge

PR #1553, opened 2025-04-29
Mep pipe size table management

### LINKED ISSUE #1552: Add Feature for Management of Pipe Sizing Data
#### Description :
Pipe sizing data often varies based on manufacturer catalogs or specific client requirements. This feature aims to standardize and simplify the management of pipe sizing datasets within a Revit template, making it easier to maintain consistency and efficiently update design data across projects.

linhnam-nguyen:

Closes #1552 

This PR introduces improvements to support management of pipe sizing data within MEP workflows:
- Pipe sizing dataset is stored in the `PipeMaterial `object
- `MEPSizes`, `Description`, `Material`, and `ScheduleType `are stored in the `CustomData `dictionary
- Full CRUD support is enabled for `MEPSizes `through the `SizeTable`

### Test files

Test File
Test Result
[PipeSegment synchronisation.xlsx](https://github.com/user-attachments/files/19960682/PipeSegment.synchronisation.xlsx)

### Changelog

- Added conversion methods: `ToRevitPipeSegment`, `PipeMaterialFromRevit`
- `PipeMaterial `is now included on Pipe objects during pull (subject to review — may not be appropriate in all cases)

### DIFF
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
@@ -71,6 +71,8 @@ namespace BH.Revit.Engine.Core
             if (type != null)
                 typeFragment = type.TypeFragmentFromRevit(settings, refObjects);
 
+            oM.MEP.System.MaterialFragments.PipeMaterial pipeMaterial = revitPipe.PipeSegment.PipeMaterialFromRevit(settings, refObjects);
+
             for (int i = 0; i < queried.Count; i++)
             {
                 BH.oM.Geometry.Line segment = queried[i];
@@ -85,6 +87,10 @@ namespace BH.Revit.Engine.Core
                 // Set the type fragment
                 if (typeFragment != null)
                     thisSegment.Fragments.Add(typeFragment);
+
+                // Set PipeMaterial fragment
+                if (pipeMaterial != null)
+                    thisSegment.CustomData.Add("PipeMaterial", pipeMaterial);
 
                 //Set identifiers, parameters & custom data
                 thisSegment.SetIdentifiers(revitPipe);
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
@@ -21,16 +21,13 @@
  */
 
 using Autodesk.Revit.DB;
+using Autodesk.Revit.DB.Plumbing;
 using BH.Engine.Adapters.Revit;
-using BH.Engine.Geometry;
-using BH.oM.Adapters.Revit;
 using BH.oM.Adapters.Revit.Settings;
 using BH.oM.Base;
 using BH.oM.Base.Attributes;
-using BH.oM.MEP.System;
 using BH.oM.MEP.System.MaterialFragments;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.ComponentModel;
 
 namespace BH.Revit.Engine.Core
@@ -46,7 +43,7 @@ namespace BH.Revit.Engine.Core
         [Input("settings", "Revit adapter settings.")]
         [Input("refObjects", "A collection of objects processed in the current adapter action, stored to avoid processing the same object more than once.")]
         [Output("pipes", "List of BHoM MEP pipes converted from a Revit pipes.")]
-        public static PipeMaterial PipeMaterialFromRevit(this Autodesk.Revit.DB.Plumbing.PipeSegment revitPipeSegment, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null)
+        public static PipeMaterial PipeMaterialFromRevit(this PipeSegment revitPipeSegment, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null)
         {
             settings = settings.DefaultIfNull();
 
@@ -61,6 +58,7 @@ namespace BH.Revit.Engine.Core
                 pipeMaterial = new PipeMaterial();
             }
 
+            Document document = revitPipeSegment.Document;
             pipeMaterial.Name = revitPipeSegment.Name;
             pipeMaterial.Roughness = revitPipeSegment.Roughness.ToSI(SpecTypeId.Length);
 
@@ -78,8 +76,8 @@ namespace BH.Revit.Engine.Core
             }
 
             pipeMaterial.CustomData.Add("Description", revitPipeSegment.Description);
-            pipeMaterial.CustomData.Add("MaterialId", revitPipeSegment.MaterialId.IntegerValue);
-            pipeMaterial.CustomData.Add("ScheduleTypeId", revitPipeSegment.ScheduleTypeId.IntegerValue);
+            pipeMaterial.CustomData.Add("Material", document.GetElement(revitPipeSegment.MaterialId).Name);
+            pipeMaterial.CustomData.Add("ScheduleType", document.GetElement(revitPipeSegment.ScheduleTypeId).Name);
             pipeMaterial.CustomData.Add("SizeTable", sizeTable);
 
             pipeMaterial.SetIdentifiers(revitPipeSegment);
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
@@ -1,0 +1,97 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2025, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using Autodesk.Revit.DB;
+using BH.Engine.Adapters.Revit;
+using BH.Engine.Geometry;
+using BH.oM.Adapters.Revit;
+using BH.oM.Adapters.Revit.Settings;
+using BH.oM.Base;
+using BH.oM.Base.Attributes;
+using BH.oM.MEP.System;
+using BH.oM.MEP.System.MaterialFragments;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Convert
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Converts a Revit PipeSegment into a BHoM PipeMaterial object, including its properties such as name, roughness, and size table. Avoids duplicate processing by utilizing a reference object collection.")]
+        [Input("revitPipeSegment", "Revit PipeSegment to be converted.")]
+        [Input("settings", "Revit adapter settings.")]
+        [Input("refObjects", "A collection of objects processed in the current adapter action, stored to avoid processing the same object more than once.")]
+        [Output("pipes", "List of BHoM MEP pipes converted from a Revit pipes.")]
+        public static PipeMaterial PipeMaterialFromRevit(this Autodesk.Revit.DB.Plumbing.PipeSegment revitPipeSegment, RevitSettings settings = null, Dictionary<string, List<IBHoMObject>> refObjects = null)
+        {
+            settings = settings.DefaultIfNull();
+
+            PipeMaterial pipeMaterial = refObjects.GetValue<PipeMaterial>(revitPipeSegment.Id);
+
+            if (pipeMaterial != null)
+            {
+                return pipeMaterial;
+            }
+            else
+            {
+                pipeMaterial = new PipeMaterial();
+            }
+
+            pipeMaterial.Name = revitPipeSegment.Name;
+            pipeMaterial.Roughness = revitPipeSegment.Roughness.ToSI(SpecTypeId.Length);
+
+            ForgeTypeId forgeTypeId = SpecTypeId.Length;
+
+            Dictionary<double, IEnumerable<double>> sizeTable = new Dictionary<double, IEnumerable<double>>();
+
+            foreach (MEPSize size in revitPipeSegment.GetSizes())
+            {
+                double nominalDiameter = size.NominalDiameter.ToSI(forgeTypeId);
+                double innerDiameter = size.InnerDiameter.ToSI(forgeTypeId);
+                double outerDiameter = size.OuterDiameter.ToSI(forgeTypeId);
+
+                sizeTable.Add(nominalDiameter, new List<double> { innerDiameter, outerDiameter});
+            }
+
+            pipeMaterial.CustomData.Add("Description", revitPipeSegment.Description);
+            pipeMaterial.CustomData.Add("MaterialId", revitPipeSegment.MaterialId.IntegerValue);
+            pipeMaterial.CustomData.Add("SizeTable", sizeTable);
+
+            pipeMaterial.SetIdentifiers(revitPipeSegment);
+
+            refObjects.AddOrReplace(revitPipeSegment.Id, pipeMaterial);
+            return pipeMaterial;
+        }
+
+        /***************************************************/
+    }
+}
+
+
+
+
+
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
@@ -79,6 +79,7 @@ namespace BH.Revit.Engine.Core
 
             pipeMaterial.CustomData.Add("Description", revitPipeSegment.Description);
             pipeMaterial.CustomData.Add("MaterialId", revitPipeSegment.MaterialId.IntegerValue);
+            pipeMaterial.CustomData.Add("ScheduleTypeId", revitPipeSegment.ScheduleTypeId.IntegerValue);
             pipeMaterial.CustomData.Add("SizeTable", sizeTable);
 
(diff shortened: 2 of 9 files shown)


Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
```diff
@@ -0,0 +1,96 @@
...
+                return pipeMaterial;
+            }
+            else
+            {
+                pipeMaterial = new PipeMaterial();
+            }
+
+            Document document = revitPipeSegment.Document;
+            pipeMaterial.Name = revitPipeSegment.Name;
+            pipeMaterial.Roughness = revitPipeSegment.Roughness.ToSI(SpecTypeId.Length);
```
pawelbaran, line 63: Are you sure that quantity of roughness is Length? I am not an expert in the field, from the memory it is unitless?

Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
```diff
@@ -86,6 +88,10 @@ public static partial class Convert
                 if (typeFragment != null)
                     thisSegment.Fragments.Add(typeFragment);
 
+                // Set PipeMaterial fragment
+                if (pipeMaterial != null)
```
pawelbaran, line 92: I would argue this should rather become a property of `PipeSectionProperty`:
https://github.com/BHoM/BHoM/blob/61c808b58d6e512d2245a8b2629cd46176781043/MEP_oM/System/SectionProperties/PipeSectionProperty.cs#L32-L33

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
```diff
@@ -0,0 +1,96 @@
...
+                double innerDiameter = size.InnerDiameter.ToSI(forgeTypeId);
+                double outerDiameter = size.OuterDiameter.ToSI(forgeTypeId);
+
+                sizeTable.Add(nominalDiameter, new List<double> { innerDiameter, outerDiameter});
+            }
+
+            pipeMaterial.CustomData.Add("Description", revitPipeSegment.Description);
+            pipeMaterial.CustomData.Add("Material", document.GetElement(revitPipeSegment.MaterialId).Name);
+            pipeMaterial.CustomData.Add("ScheduleType", document.GetElement(revitPipeSegment.ScheduleTypeId).Name);
+            pipeMaterial.CustomData.Add("SizeTable", sizeTable);
```
pawelbaran, line 81: These properties do not feel natural to me as for a material - they seem to be more system/domain-specific rather than material. However, I may be wrong taken I am not a MEP engineer 🙈 

However, I was thinking of a more elegant way of decoupling material from size table, and came up with the followning:

```C#
public class PipeSizeSet : BHoMObject
{
    public double NominalDiameter { get; set; }
    public double InnerDiameter { get; set; }
    public double OuterDiameter { get; set; }
}

public class PipeSizeTable : IFragment
{
    public string Description { get; set; }
    public List<PipeSizeSet> Table { get; set; }
}
```

...then you could query both concepts separately:
- material (with name and roughness only)
- size table (description + table)

Both of the above could be bound to `PipeSectionProperty` (assuming that all pipes of same type have same material and size table?), one via property (`Material`) and the other as a fragment added to `Fragments` property.

Then you would not need a new convert for material, just a tweak to the converts and `Update` method of `PipeSectionProperty`.

pawelbaran (commented):
Thank you @linhnam-nguyen for another fantastic contribution! I have gone through your proposal and it looks super useful, but I have 2 general comments that origin from (mostly unwritten) rules that I was trying to follow when working on Revit_Toolkit:
- we are trying to make the convert methods match concepts on both sides (in this case BHoM and Revit), so that Revit pipe converts to BHoM pipe, Revit wall to BHoM wall etc. - in your proposal BHoM pipe material is matched against Revit pipe segment, which seem not to be equivalent conceptually
- I can see you are using `CustomData` to store extra information; the idea behind `CustomData` is to enable the users store their own, problem-specific data in there rather than use it as a common platform to handle anything that does not fit anywhere else - if you are adding a new piece of extraction as a part of our core converts, the newly obtained information should have a formalised space to be passed to, e.g. a property or fragment

...having said the above, in one of my comments I proposed an alternative solution that involves adding a new pipe size table fragment and rebrands your convert methods to Query (in Revit => BHoM direction) and `Modify/Update` (BHoM => Revit).

My proposal is more of a sketch rather than a piece of code, but hopefully explains the direction I'd take well enough - please have a look and let me know how does it feel, and if it is too vague then I will try to translate it into actual code.

It is getting late here, I hope my comments will not leave you confused. If so, please shout and I will improve. Thanks!

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
linhnam-nguyen: The property appears to be unitless in the Revit API, but in the UI it is displayed as a length unit. After applying the appropriate conversion, the result aligns correctly with the BHoM output.
image

Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
linhnam-nguyen: I was confused at first as well, because the `PipeSectionProperty.PipeMaterial` is not actually of type `PipeMaterial`. After looking more closely at the modeling behavior in Revit, I realized that pipe material can also be controlled or overridden by the **plumbing system** or even by **object styles**, so it makes sense that the property isn’t strongly typed as `PipeMaterial`.

That said, the property name feels a bit misleading, which is probably what caused the confusion. I think `PipeSectionProperty.PipeMaterial`  should be renamed `PipeSectionProperty.Material` to avoid misunderstanding.

Regarding PipeMaterial, I’d love to implement it in the `FragmentSet `as well, but unfortunately, it doesn’t implement the required IFragment interface. For now, I’ve prototyped it using `CustomData`. hopefully, it can spark some consideration for a possible update to the BHoM base structure in the future.

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
pawelbaran: Hmm and are we speaking of Revit UI or yours? Trying to understand if it is a Revit glitch or maybe something on the side of BHoM or your implementation?

linhnam-nguyen: It’s how Revit displays the value. In my case, the Revit UI is set to metric, so roughness is shown in millimeters (mm). I’m not sure what it would display in Imperial, but it definitely appears as a length unit in the Revit UI.
If it weren’t a length, the unit conversion wouldn’t align properly with the BHoM output—so this seems to confirm it is treated as length at the UI level, even if it appears unitless in the API.

pawelbaran: Understood, curious if that is a Revit glitch or just difference in engineering terms - I am pretty sure roughness in STR is unitless. Regardless, happy with the explanation, all makes sense!

Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
pawelbaran: > That said, the property name feels a bit misleading, which is probably what caused the confusion. I think `PipeSectionProperty.PipeMaterial` should be renamed `PipeSectionProperty.Material` to avoid misunderstanding.

100% agreed

> Regarding PipeMaterial, I’d love to implement it in the `FragmentSet `as well, but unfortunately, it doesn’t implement the required IFragment interface. For now, I’ve prototyped it using `CustomData`. hopefully, it can spark some consideration for a possible update to the BHoM base structure in the future.

`PipeMaterial` implements `IMaterialProperties` interface, which can be stored under `Properties` prop of `Material` class:
https://github.com/BHoM/BHoM/blob/61c808b58d6e512d2245a8b2629cd46176781043/Physical_oM/Materials/Material.cs#L48-L49

The concept behind it is to enable physical material to carry various pieces of domain-specific information (e.g. mechanical and thermal in case of structural analysis etc.). The easiest way to leverage it in your case would be:
- create physical `Material` with `Name` equal to the pipe material
- create `PipeMaterial` with `Name` and `Roughness`
- add `PipeMaterial` to properties of `Material`
- set `PipeMaterial` property of `PipeSectionProperty` (again, valid observation about naming of the property)

Would that make sense?

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
linhnam-nguyen: Your proposal is clean and makes sense from a data modeling perspective. That said, I chose the current approach to align closely with how Autodesk structures MEP modeling. 

To my knowledge, a `PipeSegment `that contains pipe size data must be associated with a specific combination of material and schedule, as this unique pairing is required to access the` ICollection<MEPSize> `data. This means that every pipe segment has a base material defined at the time of modeling, even though it can later be overridden by the system or object styles, as I mentioned earlier.

In the context of plumbing sizing, the only material-related property typically considered is interior roughness, which is essential for pressure drop calculations. For that reason, I believe it's reasonable to follow Autodesk’s structure as-is.

For domain-specific properties, I think it's appropriate to implement domain materials in `BH.oM.MEP.System.MaterialFragments`. For other Autodesk domain-specific inputs, it might be better to add them as part of the FragmentSet of the respective domain material.

That said, without a proper conversion for PipeSegment, we wouldn’t be able to synchronize the design data as the intention of my PR.

Here’s a quick summary from my understanding of how sizing data is handled across systems:

- PipeSegment → Segment → ICollection<MEPSize> → full sizing data
- CableTraySizes → IEnumerable<MEPSize> → uses only NominalDiameter
- ConduitSizes → IEnumerable<ConduitSize> → similar to MEPSize with extra BendRadius property
- CableSize → accessible from Revit 2026+
- DuctSizeSettings → IEnumerable<KeyValuePair<DuctShape, DuctSizes>> → DuctSizes: IEnumerable<MEPSize> → rarely need updates

For a more generalized future-facing approach, I agree that we could abstract PipeSizeTable into something like a SizeSet. However, I’d argue that we keep the initial implementation simple. One possible idea is to include a SizeSet dictionary as a fragment on an IMEPMaterial interface, avoiding further fragmentation in BHoM base:

```
public interface IMEPMaterial : IBHoMObject
{
    Dictionary<double, List<double>> SizeSet;
}
```
This way, we can support broader generalization without overcomplicating the core structure.

Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
linhnam-nguyen: > - create physical Material with Name equal to the pipe material
> - create PipeMaterial with Name and Roughness
> - add PipeMaterial to properties of Material
> - set PipeMaterial property of PipeSectionProperty (again, valid observation about naming of the property)

Would it be okay if I slightly modified your mapping as follows?

- Create physical `Material` with the Name equal to the value of the `RBS_PIPE_MATERIAL_PARAM` parameter
- Create `PipeMaterial` with Name and Roughness corresponding to `PipeSegment.Name` and `PipeSegment.Roughness`
- (then continue as yours)

Because `PipeMaterial ` will include `ScheduleType ` name.

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
pawelbaran: I think I am starting to understand the complexities involved - apologies for my ignorance but I am none of a MEP engineer. Zooming out, I think my main concern is that `BH.oM.MEP.Pipe` and all its properties are core BHoM objects i.e. they need to be modelled in a software-agnostic way, while the concept of Revit sizing seems very specific to Revit itself. Therefore, I can't agree with this part:

> For a more generalized future-facing approach, I agree that we could abstract PipeSizeTable into something like a SizeSet. However, I’d argue that we keep the initial implementation simple. One possible idea is to include a SizeSet dictionary as a fragment on an IMEPMaterial interface, avoiding further fragmentation in BHoM base:
> 
> ```
> public interface IMEPMaterial : IBHoMObject
> {
>     Dictionary<double, List<double>> SizeSet;
> }
> ```
> 
> This way, we can support broader generalization without overcomplicating the core structure.

`IMEPMaterial` needs to remain software-agnostic, i.e. it should not be mixed with sizing, even if Revit does it. I'd rather turn sizing info into an `IFragment` that would be bound to `Pipe` object at some level (either of the list: pipe itself, its section property, material of section property, material fragment). Thanks to that, we'd retain the core oM in the original shape, but enable working with sizing information in Revit workflows.

Also, as mentioned before, I do not think putting equality sign between BHoM `PipeMaterial` and Revit `PipeSegment` is correct - although they map well in your particular workflow, they are fundamentally different concepts, i.e. I would avoid creating `Convert` methods for the pair, in favour of `Query` method that would extract the sizing table (same as `PipeSectionProperty` is queried from `Revit` pipe).

```C#
public static `PipeSizeTable` PipeSizeTable(this PipeSegment pipeSegment)
{
    (...)
}
```

Trying to think how to combine your domain knowledge with my general feel for the oM. Again, bringing up my proposal for the new classes, which could sit in Revit_Toolkit rather than BHoM:

```C#
public class PipeSizeSet : BHoMObject
{
    public double NominalDiameter { get; set; }
    public double InnerDiameter { get; set; }
    public double OuterDiameter { get; set; }
}

public class PipeSizeTable : IFragment
{
    public string Description { get; set; }
    public List<PipeSizeSet> Table { get; set; }
}
```

Would it be possible to do the following:
- on `Pull`, call `Query.PipeSizeTable` method against `Pipe.PipeSegment` and attach the output `PipeSizeTable` fragment to one of the following:
    a) `Pipe` 
    b) `PipeSectionProperty`
    c) `Material`
    d) `PipeMaterial`
    ...so that the BHoM representation of a pipe matches its Revit equivalent? You tell me which level is most appropriate 😉 
- on `Push`, tweak `BH.Revit.Engine.Core.Modify.Update(Pipe)` and `BH.Revit.Engine.Core.Convert.ToRevitPipe` methods in a way that would read the fragment and update Revit `PipeSegment` correspondent to pipe accordingly?

If so, then I would opt for such solution. In simple terms: we'd add a fragment in the Revit oM namespace and make sure that it is being handled on Revit Push/Pull independently of the core BHoM concepts that are much more rigid. Smallest and simplest I could think of...

Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
pawelbaran: Sure, that could work! Let's just make sure that schedule type and sizing table are stored in fragments: either one (if both pieces of information are bound to each other) or two separate, to avoid Revit-specific concepts in the core oM

Revit_Core_Engine/Convert/MEP/FromRevit/PipeMaterial.cs
linhnam-nguyen: You're right that these sizing concerns should remain Revit-specific, and I agree—it makes sense to have a PipeSizeTable fragment attached to PipeMaterial.

However, I think `PipeMaterial `should have its own dedicated conversions, allowing it to be pulled directly from RevitDBType rather than being accessed through the `PipeSectionProperty`. This should'nt be related to pipe instance, since deeper calculations could introduce unnecessary processing overhead. 

As these classe are RevitToolkit specific, I suggest mapping properly as below, thanks for your clarification about `FragmentSet ` in `PipeMaterial ` :
```c#
public class PipeSize : BHoMObject
{
    public double InnerDiameter { get; set; }
    public double OuterDiameter { get; set; }
}

public class PipeSizeTable : IFragment 
{
    public string ScheduleType { get; set; } // as this fragment is Revit specific
    public string Description { get; set; }
    public Dictionary<double, PipeSize> PipeSizes  { get; set; } // allow no duplicate sizes
}
``` 

Although it's not a direct one-to-one mapping, `PipeMaterial ` is the most suitable candidate to align with Revit's PipeSegment. With the `PipeSizeTable `fragment within `PipeMaterial`, we can keep BHoM base clean, as originally intended.

When I thought again about workflows involving `PipeSizeTable `, the focus was more on standardization than on detailed model inspection—so having it decoupled from pipe instance seems more appropriate.

pawelbaran: OK, I am getting increasingly convinced 👍 Just looked up the data structure in Revit, and it actually looks like their naming is a bit unfortunate: in my simplistic mind _segment_ is a piece of a bigger system, i.e. it would be unique per each element (or even multiple segments per a single pipe, assuming there are other elements connected along it). Meanwhile:

image

So yeah, mapping of `PipeMaterial` to `PipeSegment` does not look great, but seems to have sense 👍

One more question: shouldn't `ScheduleType` property be of type `int` to map by `ElementId`?

linhnam-nguyen: To my understanding, ScheduleType represents the manufacturing standard, which can vary across suppliers and markets. When prototyping, I was initially unsure whether to use the ElementId or the string name.

In the end, I chose to use the name for a readable output, as it’s more accessible to regular users who typically don’t have easy access to ElementIds. That said, the `ElementId `can still be retrieved directly when needed using:

```c#
public static ElementId GetPipeScheduleId(
	Document doc,
	string name
)
``` 

With this in mind, would you prefer to take the lead on polishing the implementation, or should I proceed to develop the detailed logic for creating the Revit objects and mapping as discussed?

pawelbaran: Ah, if mapping by name is possible then probably best to stick to the string.

Concerning who does it - more than happy for you to take the lead, I am currently snowed under other threads unfortunately...

linhnam-nguyen:
Hello @pawelbaran, 

I've refactored the code to adapt the structure as we discussed.
You can find the sample application in my new test file here:

[250515 - PipeSegment.synchronisation.xlsx](https://github.com/user-attachments/files/20229173/250515.-.PipeSegment.synchronisation.xlsx)

I’d love to hear your thoughts on these changes.

pawelbaran (commented):
I started reading the code and making comments, but at some point I felt like I need to confirm my thoughts with code implementation, which at some point led me to branching off from your branch to propose an alternative solution - see https://github.com/BHoM/Revit_Toolkit/pull/1563.

Key changes I have made:
- since Revit pipe segment is set per item same as pipe diameter, it all gets converted separately per each instance: the call stack is as follows: `PipeFromRevit` `=>PipeSectionProperty` => `PipeMaterialFromRevit`, pipe design data being extracted at the last step
- following the above, nesting in the oM after pull is as follows: `BH.oM.MEP.System.Pipe` => `SectionProperty` => `PipeMaterial` => `Properties[0]` => `Fragment` of type `PipeDesignData`
- on `Push` we have 2 paths:
    a) Create: `ToRevitPipe` and `ToRevitPipeSegement` (if needed) - both only create new Revit elements, never update existing ones
    b) Update: `Update` methods against `Pipe` or `PipeSegment`, depending on whether the pushed object is `Pipe` or `PipeMaterial`, accordingly

Everything else is minor tweaks and refactors, which hopefully make sense. I also hope you do not mind me taking over your work - it just felt much quicker for me to simply commit my changes rather than trying to explain all of them via text comments. Naturally, if you disagree with that step or with changes made, please shout 👍 Thanks!

PR #1563, opened 2025-05-28
Mep pipe size table management finalised

### LINKED ISSUE #1552: Add Feature for Management of Pipe Sizing Data

pawelbaran:

Closes #1552

This is a finalised version of https://github.com/BHoM/Revit_Toolkit/pull/1553 - picked up from @linhnam-nguyen as a result of an extended code review during which I did most of the coding to prove to myself the comments I was making.

### Test files

Grasshopper script on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FRevit%5FToolkit%2F%231563%2DMEP%5FPipeSizeTableManagement).

### DIFF
--- a/Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
+++ b/Revit_Core_Engine/Convert/MEP/FromRevit/Pipe.cs
@@ -71,6 +71,8 @@ namespace BH.Revit.Engine.Core
             if (type != null)
                 typeFragment = type.TypeFragmentFromRevit(settings, refObjects);
 
+            oM.MEP.System.MaterialFragments.PipeMaterial pipeMaterial = revitPipe.PipeSegment.PipeMaterialFromRevit(settings, refObjects);
+
             for (int i = 0; i < queried.Count; i++)
             {
                 BH.oM.Geometry.Line segment = queried[i];
@@ -85,6 +87,10 @@ namespace BH.Revit.Engine.Core
                 // Set the type fragment
                 if (typeFragment != null)
                     thisSegment.Fragments.Add(typeFragment);
+
+                // Set PipeMaterial fragment
+                if (pipeMaterial != null)
+                    thisSegment.CustomData.Add("PipeMaterial", pipeMaterial);
 
                 //Set identifiers, parameters & custom data
                 thisSegment.SetIdentifiers(revitPipe);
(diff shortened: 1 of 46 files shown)


pawelbaran:
Looks like there is 2 against one on `NominalDiameter` property @linhnam-nguyen 🙈 I updated the code following the suggestion from @michal-pekacki, which I also believe is more appropriate. Hopefully that will also work for you 👍

linhnam-nguyen:
@pawelbaran, otherwise than the changes to address `NominalDiameter`, your modifications are great. I think we should add a check to ensure the uniqueness of NominalDiameter to prevent any bugs. Sorry for trying to package everything into the conversion methods—it made the code a bit more complicated than necessary. Thanks for your refactoring work!

michal-pekacki:
> > Looks like there is 2 against one on `NominalDiameter` property @linhnam-nguyen 🙈 I updated the code following the suggestion from @michal-pekacki, which I also believe is more appropriate. Hopefully that will also work for you 👍
> 
> Hi @michal-pekacki and @pawelbaran, The reason I use a dictionary on NominalDiameter is because Revit doesn’t allow multiple NominalDiameter values in a PipeSegment schedule. While it would make more sense from a data perspective to include NominalDiameter in MEPSize—which would let us have a richer SizeSet with different ID and OD for the same ND—Revit’s constraints prevent us from doing that. So, I use the dictionary as a precheck to ensure the SizeSet sent to Revit is valid.

Yes, from PipeSegment perspective dictionary makes perfect sense, so I don't have strong opinion here, but I feel `SizeSet` can be used in other cases, and it would be clearer to base on properties.

pawelbaran:
@linhnam-nguyen you are right from the classic programming perspective, 100% agree with your arguments. However, as you have probably noticed BHoM class definitions are just schema definitions: only get/set properties essentially. This means that we inherently allow the users build invalid objects if they really like, assuming that garbage in == garbage out. I would be fine with taking similar approach in our current case (duplicate nominal diameters), in favour of the three values (nominal, inner and outer) being bundled together as one conceptual entity.

I have just added a validation method that culls duplicates and makes sure that all inputs are valid - hopefully that will be enough for your use case, if not then please shout and we'll improve it further 👍

pawelbaran:
@BHoMBot check versioning
@BHoMBot check installer

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check required

michal-pekacki (approved):
Happy to approve after testing on given model and GH script 👍

michal-pekacki:

PR #1579, opened 2025-06-18
Add full Pile conversion support

### LINKED ISSUE #1578: Add Push/Pull functionality for Pile objects in the Physical
#### Description:

Add functionality to be able to push and pull `Pile` objects in to Revit.

peterjamesnugent:

Closes #1578 

### Test files

[Revit Model and GH Script](https://burohappold.sharepoint.com/:f:/s/BHoM/EiFb9JgHulhEoVSYDokSaNYBrQYIEeVTw98V3xnGWbRjtw?e=7DbRUe)

#### New Features
- Added full bidirectional conversion support for `BH.oM.Physical.Elements.Pile` elements;
- Implemented point-based family instance placement for pile foundations;
- Created pile-specific profile extraction with multiple fallback strategies;
- Added intelligent profile detection from parameter names and family naming conventions.

#### Enhanced Methods
- Extended `BHoMType` recognition to support Pile elements via built-in categories and `StructuralType.Footing`
- Added `LocationCurvePile` method for pile-specific geometry extraction
- Created `PileFramingElementProperty` method with enhanced profile extraction capabilities

### DIFF
--- a/Revit_Core_Engine/Compute/GenerateProfile.cs
+++ b/Revit_Core_Engine/Compute/GenerateProfile.cs
@@ -211,11 +211,12 @@ namespace BH.Revit.Engine.Core
                     t.Commit();
                 }
 
-                string tempLocation = SanitizePath($"{m_TempFolder}\\{element.ProfileFamilyName()}.rfa");
+                string tempFolder = Path.GetTempPath();
+                string tempLocation = SanitizePath($"{tempFolder}\\{element.ProfileFamilyName()}.rfa");
                 try
                 {
-                    if (!System.IO.Directory.Exists(m_TempFolder))
-                        System.IO.Directory.CreateDirectory(m_TempFolder);
+                    if (!Directory.Exists(tempFolder))
+                        Directory.CreateDirectory(tempFolder);
 
                     SaveAsOptions saveOptions = new SaveAsOptions();
                     saveOptions.OverwriteExistingFile = true;
@@ -223,7 +224,7 @@ namespace BH.Revit.Engine.Core
                 }
                 catch (Exception ex)
                 {
-                    BH.Engine.Base.Compute.RecordError($"Creation of a freeform Revit profile geometry failed because the family could not be temporarily saved in {m_TempFolder}. Please make sure the folder exists and you have access to it, then try to empty it in case the issue persists.");
+                    BH.Engine.Base.Compute.RecordError($"Creation of a freeform Revit profile geometry failed because the family could not be temporarily saved in {tempFolder}. Please make sure the folder exists and you have access to it, then try to empty it in case the issue persists.");
                     familyDocument.Close(false);
                     return null;
                 }
@@ -232,7 +233,7 @@ namespace BH.Revit.Engine.Core
 
                 try
                 {
-                    System.IO.File.Delete(tempLocation);
+                    File.Delete(tempLocation);
                 }
                 catch (Exception ex)
                 {
@@ -708,7 +709,6 @@ namespace BH.Revit.Engine.Core
         /***************************************************/
 
         private static readonly string m_FamilyDirectory = @"C:\ProgramData\BHoM\Resources\Revit\Families";
-        private static readonly string m_TempFolder = @"C:\temp";
 
         private static readonly Dictionary<Type, string> m_FamilyFileNames = new Dictionary<Type, string>
         {


Revit_Core_Engine/Query/LocationCurve.cs
```diff
@@ -219,6 +221,74 @@ public static BH.oM.Geometry.Line LocationCurveColumn(this FamilyInstance family
...
+                double startExtension = extensions.Item1;
+                double endExtension = extensions.Item2;
+
+                if (Math.Abs(startExtension) > settings.DistanceTolerance || Math.Abs(endExtension) > settings.DistanceTolerance)
+                {
+                    Vector direction = curve.Direction();
+                    curve = new oM.Geometry.Line { Start = curve.Start - direction * startExtension, End = curve.End + direction * endExtension };
+                }
+            }
+            else
```
pawelbaran, line 272: I do not think this code will ever be hit

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -145,8 +145,109 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
+                return familyInstance;
+
+            settings = settings.DefaultIfNull();
+
+            BH.oM.Geometry.Line locationLine = framingElement.PileLine();
+            if (locationLine == null)
+                return null;
+
+            Level level = document.LevelBelow(locationLine, settings);
+            if (level == null)
```
pawelbaran, line 162: Hmm I would rather look for closest level above (with potential warning) rather than immediately returning `null`

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -145,8 +145,109 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
+            FamilySymbol familySymbol = framingElement.ElementType(document, settings);
+            if (familySymbol == null)
+            {
+                Compute.ElementTypeNotFoundWarning(framingElement);
+                return null;
+            }
+
+            FamilyPlacementType familyPlacementType = familySymbol.Family.FamilyPlacementType;
+            // Accept point-based placement types for piles instead of curve-based
+            if (familyPlacementType != FamilyPlacementType.OneLevelBased &&
```
pawelbaran, line 178: Hmm from what I know piles are usually set as `FamilyPlacementType.OneLevelBased`, not sure if we should try to support anything else 'in advance'

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -145,8 +145,109 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
+                        if (param != null && param.HasValue && !param.IsReadOnly)
+                            familyInstance.StructuralMaterialId = material.Id;
+                        else
+                            BH.Engine.Base.Compute.RecordWarning(string.Format("The BHoM material has been correctly converted, but the property could not be assigned to the Revit element. ElementId: {0}", familyInstance.Id));
+                    }
+                }
+            }
+
+            // Set pile height constraint using the height from the original line geometry
+            double pileHeight = Math.Abs(locationLine.End.Z - locationLine.Start.Z);
```
pawelbaran, line 213: This needs to be converted `FromSI`

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -145,8 +145,109 @@ public static FamilyInstance ToRevitFamilyInstance(this Column framingElement, D
...
+                topOffsetParam.Set(pileHeight);
+                heightSet = true;
+            }
+            
+            // Set base level offset to zero (pile starts at base level)
+            Parameter baseOffsetParam = familyInstance.get_Parameter(BuiltInParameter.FAMILY_BASE_LEVEL_OFFSET_PARAM);
+            if (baseOffsetParam != null && !baseOffsetParam.IsReadOnly)
+            {
+                baseOffsetParam.Set(0.0);
+            }
```
pawelbaran, line 239: That does not work at all actually. Based on my knowledge, what needs to be done is:
1. Set the top level param (`BuiltInParameter.FAMILY_LEVEL_PARAM`) to the host level
2. Set the offset from that level to top of the pile (`BuiltInParameter.FLOOR_HEIGHTABOVELEVEL_PARAM`)
3. Set the pile length, which is a bit tricky, e.g. in the test file the parameter is not a system one, just user-created _BHE_Pile Depth_ - we somehow need to know that it exists under this name and not the other. To make it more fun, not sure if our physical length should not be correspondent to _BHE_Pile Length_ (which is read only, derived from _BHE_Pile Depth_, see snip below).

    image

    That becomes tricky to the point where I would argue that the best solution would be to put a pile family under _C:\ProgramData\BHoM\Resources\Revit\Families_ and hardcode the parameter names in the code, effectively limiting support to a single family, with potential to extend using `settings.MappingSettings` to map across other parameter names.

Please let me know in case that's too confusing @peterjamesnugent, happy to support/contribute. Cheers!

pawelbaran (changes requested):

pawelbaran:
@BHoMBot check compliance
@BHoMBot check versioning

pawelbaran:
@BHoMBot check versioning

peterjamesnugent:
@pawelbaran can't seem to add my review (probably because I created the PR). In terms of what you have committed:
I have tested with Revit 2022.1 and Revit 2024 (that's what I have installed):
- `Pull` works for default profiles in Revit; 
- `Update` works for different profiles and is reflected in the UI;
- `Push` works for different profiles;

I did get an error when pushing `FreeFormProfile` with rotation but that's because `Rectangular Piles.rfa` was saved by a higher version (so nothing we can do about that and it's a clear error that the user can rectify). Though it did still push the `Pile` with Rectangular Piles 800x400mm.

Also, when using a custom name in the `FramingProperty.Name` if there's a space at the front i.e. " ABC 123" is raises an illegal character error - so there's likely a line somewhere we need to add `Trim()` when processing the name property.

PR #1583, opened 2025-08-06
`LinkedElementsInView` - spatial elements fix

michal-pekacki:

Closes #

`ElementIntersectsSolidFilter` doesn't work for `SpatialElements` and they need to be captured by checking the intersection between location point and view solid.


pawelbaran:
@BHoMBot check ready-to-merge

PR #1616, opened 2025-11-10
Refresh button icons

vietle-bh:

I updated icons for the ribbon buttons as proposed by @JanineConway 👍 

### Test files

Simply check if the right icons show up on the ribbon.


JanineConway (approved):
All showing well

PR #1651, opened 2026-03-31
Push of pad foundation 

WojciechBuda00:

Closes #1650 

### Test files

[installer](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Revit_Toolkit/%231651-SupportRevitPadFoundation_Push?csf=1&web=1&e=HAUSDG)


WojciechBuda00:
@BHoMBot check installer

Revit_Core_Engine/Compute/GeneratePadFoundation.cs
```diff
@@ -0,0 +1,222 @@
...
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Generates a Revit family type (and its parent family, if not loaded yet) that represents the profile of a given BHoM pad foundation." +
+                     "\nThe profile is created based on the template families stored in C:\\ProgramData\\Resources\\Revit.")]
```
pawelbaran: Not sure if I would put this information here, rather not relevant to the user.

Revit_Core_Engine/Compute/GeneratePadFoundation.cs
```diff
@@ -0,0 +1,222 @@
...
+        [Input("padFoundation", "BHoM pad foundation to generate the Revit profile for.")]
+        [Input("document", "Revit document, in which the family type will be created.")]
+        [Input("settings", "Settings to be used when generating the family type.")]
+        [Output("symbol", "Created Revit family type that represents the profile of the input BHoM pad foundation.")]
+        public static FamilySymbol GeneratePadFoundation(this PadFoundation padFoundation, Document document, RevitSettings settings = null)
```
pawelbaran: I would consider renaming this method to `GeneratePadFoundationType`, this is what this method ultimately does

Revit_Core_Engine/Compute/GeneratePadFoundation.cs
```diff
@@ -0,0 +1,222 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Generates a Revit family type (and its parent family, if not loaded yet) that represents the profile of a given BHoM pad foundation." +
+                     "\nThe profile is created based on the template families stored in C:\\ProgramData\\Resources\\Revit.")]
+        [Input("padFoundation", "BHoM pad foundation to generate the Revit profile for.")]
+        [Input("document", "Revit document, in which the family type will be created.")]
+        [Input("settings", "Settings to be used when generating the family type.")]
+        [Output("symbol", "Created Revit family type that represents the profile of the input BHoM pad foundation.")]
```
pawelbaran:
```suggestion
        [Output("symbol", "Created Revit family type that represents the outline of the input BHoM pad foundation.")]
```

Revit_Core_Engine/Compute/GeneratePadFoundation.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            return result;
+        }
+
+        /***************************************************/
+
+        private static Family GenerateFreeformPadFoundation(this Document document, PadFoundation padFoundation, string familyName, RevitSettings settings = null)
+        {
+            throw new NotImplementedException("GenerateFreeformPadFoundation is not implemented.");
+        }
```
pawelbaran: I think I would delete this method and simply return null in the caller method

Revit_Core_Engine/Compute/GeneratePadFoundation.cs
```diff
@@ -0,0 +1,222 @@
...
+            else
+                return name;
+        }
+
+        /***************************************************/
+
+        private static void ICopyFoundationDimensions(this PadFoundation padFoundation, FamilySymbol targetSymbol, RevitSettings settings = null)
+        {
+            CopyFoundationDimensions(padFoundation as dynamic, targetSymbol, settings);
+        }
```
pawelbaran: No need for an I method here, there won't be more than one pad foundation type, you could remove this and simply call `CopyFoundationDimensions` where needed

Revit_Core_Engine/Modify/SetLocation.cs
```diff
@@ -344,6 +344,25 @@ public static bool SetLocation(this FamilyInstance element, Pile pile, RevitSett
 
         /***************************************************/
 
+        [Description("Sets the location of a given Revit FamilyInstance based on a given BHoM PadFoundation.")]
+        [Input("element", "Revit FamilyInstance to be modified.")]
+        [Input("padFoundation", "BHoM PadFoundation acting as a source of information about the new location.")]
+        [Input("settings", "Revit adapter settings to be used while performing the operation.")]
+        [Output("success", "True if location of the input Revit FamilyInstance has been successfully set.")]
+        public static bool SetLocation(this FamilyInstance element, PadFoundation padFoundation, RevitSettings settings)
```
pawelbaran, line 353: You are not updating the rotation by the looks of it, also would be great to enable updating the dimensions - most similar case I can think of is updating the profile of a framing element (please have a look how is it being done)

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+    {
+
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Extracts the outer rectangular boundary of a PadFoundation as a Polyline.")]
+        [Input("element", "PadFoundation element whose boundary should be extracted.")]
+        [Output("outline", "Polyline representing the PadFoundation external boundary.")]
+        public static Polyline ExtractBoundaryFoundationGeometry(this PadFoundation element)
```
pawelbaran: This could simply be `Query.Boundary` or similar

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            if (element.Location == null)
+                return null;
+
+            if (element.Location.InternalBoundaries.Count != 0)
+            {
+                BH.Engine.Base.Compute.RecordError("InternalBoundaries error");
```
pawelbaran: Error message could be improved

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+
+            List<ICurve> segments = element.Location.ExternalBoundary.ISubParts().ToList();
+            if (segments.Any(x => !(x is BH.oM.Geometry.Line)))
+            {
+                BH.Engine.Base.Compute.RecordError("");
```
pawelbaran: ...

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+                return false;
+
+            var points = polyline.ControlPoints;
+            int pointCount = points.Count;
+
+            //Reecrtangle check  closing point
+            if (pointCount != 5)
+                return false;
+
+            List<BH.oM.Geometry.Line> edges = new List<BH.oM.Geometry.Line>();
```
pawelbaran: Do you need to create the edges here, isn't it enough to check if len(cpts[2] - cpts[0]) == len(cpts[3] - cpts[1]) within tolerance?

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            return BH.Engine.Geometry.Create.Polyline(segments.Cast<BH.oM.Geometry.Line>().ToList());
+        }
+
+        /***************************************************/
+
+        [Description("Checks whether a Polyline represents a rectangle.")]
+        [Input("polyline", "Polyline to check.")]
+        [Output("isRectangular", "True if the polyline represents a rectangle; otherwise false.")]
+        public static bool IsRectangularFoundationGeometry(this Polyline polyline)
```
pawelbaran: This could simply be named `IsRectangular`

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            return Math.Abs(d1 - d2) <= tolerance;
+        }
+
+        /***************************************************/
+
+        [Description("Computes the rotation angle of a local coordinate system in the XY plane.")]
+        [Input("localCS", "Local coordinate system to compute the rotation angle for.")]
+        [Output("angle", "Rotation angle in radians.")]
+        public static double ComputeRotationAngleFoundationGeometry(this Cartesian localCS)
```
pawelbaran: Could be just RotationAngleInXY, starting with a check if local Z is parallel to global Z

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            return Math.Atan2(global.CrossProduct(local).Z, global.DotProduct(local));
+        }
+
+        /***************************************************/
+
+        [Description("Computes the total thickness (sum of construction layers) of a PadFoundation.")]
+        [Input("element", "PadFoundation element whose thickness should be computed.")]
+        [Output("thickness", "Total thickness of all construction layers.")]
+        public static double GetThicknessFromConstrFoundationGeometry(this PadFoundation element)
```
pawelbaran: I'd expect to find something like `Thickness(this Construction)` or `TotalThickness(this Counstruction)` in Physical_Engine. Then instead of writing all of that you could do something like `var thk = element.Construction?.Thickness()`. This feels a lot like reinventing the wheel - and if it is not, then I'd add the method mentioned above to Physical_Engine and call it here

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+            return totalDepth;
+        }
+
+        /***************************************************/
+
+        [Description("Gets width and length dimensions of a rectangular Polyline.")]
+        [Input("polyline", "Polyline representing a rectangle.")]
+        [Output("dimensions", "Tuple containing (width, length) in the same units as the input polyline.")]
+        public static (double width, double length) GetRectangleDimensionsFoundationGeometry(this Polyline polyline)
```
pawelbaran:
```suggestion
        public static (double width, double length) RectangleDimensions(this Polyline polyline)
```

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            if (polyline == null || polyline.ControlPoints == null)
+            {
+                BH.Engine.Base.Compute.RecordWarning("Cannot get rectangle dimensions: polyline or control points are null.");
```
pawelbaran: Nothing to return here?

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        /***************************************************/
+
+        [Description("Gets width and length dimensions of a rectangular Polyline.")]
+        [Input("polyline", "Polyline representing a rectangle.")]
+        [Output("dimensions", "Tuple containing (width, length) in the same units as the input polyline.")]
+        public static (double width, double length) GetRectangleDimensionsFoundationGeometry(this Polyline polyline)
+        {
+            if (polyline == null || polyline.ControlPoints == null)
```
pawelbaran: Here I'd simply start from `if (!IsRectangular(polyline))` to eliminate all non-rectangles, then go directly to returning lengths of the first two edges

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+            return (width, length);
+        }
+
+        /***************************************************/
+
+        [Description("Loads the PadFoundation rectangular family symbol template (or activates an existing one).")]
+        [Input("document", "Revit document where the family symbol should be loaded/activated.")]
+        [Input("settings", "Revit adapter settings used for loading the family symbol.")]
+        [Output("symbol", "FamilySymbol for the rectangular PadFoundation template, or null if not found.")]
+        public static FamilySymbol LoadPadRectangleTemplateFoundationGeometry(this Document document, RevitSettings settings)
```
pawelbaran: I would rename it and move to `Compute` class

```suggestion
        public static FamilySymbol LoadPadRectangleTemplate(this Document document, RevitSettings settings)
```

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        /***************************************************/
+
+        [Description("Generates (duplicates and configures) a PadFoundation FamilySymbol type based on foundation dimensions.")]
+        [Input("element", "PadFoundation element whose dimensions will drive the generated FamilySymbol type parameters.")]
+        [Input("document", "Revit document where the symbol type will be created/duplicated.")]
+        [Input("settings", "Revit adapter settings used while generating the symbol type.")]
+        [Output("symbol", "Generated FamilySymbol with parameters matching the foundation dimensions, or null on failure.")]
+        public static FamilySymbol GenerateFoundationTypeFoundationGeometry(this PadFoundation element, Document document, RevitSettings settings = null)
```
pawelbaran: Isn't that a duplicate of `Comupte.GeneratePadFoundation`??

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            return foundationSymbol;
+        }
+
+        /***************************************************/
+
+        [Description("Gets the origin point (centroid) of a PadFoundation.")]
+        [Input("element", "PadFoundation element whose origin should be computed.")]
+        [Output("origin", "Origin point (centroid) of the foundation external boundary, or null if invalid.")]
+        public static BH.oM.Geometry.Point GetFoundationOriginFoundationGeometry(this PadFoundation element)
```
pawelbaran:
```suggestion
        public static BH.oM.Geometry.Point Origin(this PadFoundation element)
```

pawelbaran (changes requested):
There is still quite some work to do, see my comments. On top of that, I've tested the PR using the foundations pulled from https://github.com/BHoM/Revit_Toolkit/pull/1646, and have a few comments comments:
- one element comes out rotated by 90 degrees
- foundations with nonlinear edges throw nasty errors on push
    Image
- have not tested, but what will happen with foundations with outlines as non-rectangular polylines?

Image

We also need to catch up on 2 more topics:
- setting up the families
- update of foundation dimensions on `Update`

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,175 @@
...
+            return element?.Outline()?.Centroid();
+        }
+
+        /***************************************************/
+
+        [Description("Checks whether a Polyline represents a rectangle.")]
+        [Input("polyline", "Polyline to check.")]
+        [Input("settings", "Revit adapter settings containing tolerance values. If null, default tolerance will be used.")]
+        [Output("isRectangular", "True if the polyline represents a rectangle; otherwise false.")]
+        public static bool IsRectangle(this Polyline polyline, RevitSettings settings = null)
```
pawelbaran: Please port it to BHoM_Engine/Geometry_Engine as a `Query` method - you need to create a sister branch in BHoM_Engine repo with the same name as this and raise a PR in there as well.

When porting, need to change 2nd method parameter from `RevitSettings` to `double` representing tolerance

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            List<Point> pts = polyline?.ControlPoints;
+            if (pts == null || pts.Count < 4)
```
pawelbaran: Should be 3 criteria:
- number of control points == 5
- `polyline.IsClosed(tolerance) == true`
- both diagonals the same within tolerance

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,175 @@
...
+            double diagonal2 = pts[3].Distance(pts[1]);
+
+            return Math.Abs(diagonal1 - diagonal2) <= tol;
+        }
+
+        /***************************************************/
+        /****              Private methods              ****/
+        /***************************************************/
+
+        private static Vector LongestEdgeInXY(this Polyline polyline)
```
pawelbaran: Need to pass tolerance to this method, also could move it as a more general method to BHoM_Engine:
```suggestion
        private static Vector LongestEdgeDirection(this Polyline polyline, double tolerance)
```

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            List<Point> pts = polyline?.ControlPoints;
+            if (pts == null || pts.Count < 2)
+                return null;
+
+            int n = pts.Count;
```
pawelbaran: This method was written by AI and seems quite overcooked. What needs to be done is:
- create `Dictionary<Vector, double>` storing vectors and their correspondent total lengths of parallel edges
- for each edge (`polyline.SubParts()`) find direction and length, then check if dir is parallel is to any key in the dictionary, if yes then add length to that pair, otherwise create a new one
- return key of the largest value

This will find direction with longest sum of paralellel edge lengths

Revit_Core_Engine/Query/IsMatchingOutline.cs
```diff
@@ -0,0 +1,107 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Checks whether the given BHoM Polyline outline matches the outline of the family freeform extrusion in Revit.")]
+        [Input("family", "Revit family whose extrusion outline should be compared.")]
+        [Input("orientedOutline", "BHoM polyline outline to match against the family outline.")]
+        [Input("settings", "Revit adapter settings providing distance tolerance for the matching.")]
+        [Output("matches", "True if the family extrusion outline matches the input outline.")]
+        public static bool IsMatchingOutline(this Family family, Polyline orientedOutline, RevitSettings settings)
```
pawelbaran: Also worth tweaking the descriptions to be more specific

```suggestion
        public static bool IsMatchingOutline(this Family padFoundationFamily, Polyline orientedOutline, RevitSettings settings)
```

Revit_Core_Engine/Modify/OrientToOrigin.cs
```diff
@@ -0,0 +1,27 @@
...
+        [Description("Transforms a given Polyline into a standardized orientation by translating it to the origin and rotating it in XY plane.")]
+        [Input("outline", "Polyline to be oriented to the origin.")]
+        [Output("orientedOutline", "Oriented polyline translated to the origin and rotated in XY plane.")]
+        public static Polyline OrientToOrigin(this Polyline outline)
+        {
+            (Vector translation, double rotation) = outline.TransformToOriginInXY();
+            if (translation == null || double.IsNaN(rotation))
+                return null;
+
+            Vector moveToOrigin = (new Point() - outline.Centroid()).ProjectToXY();
```
pawelbaran: This line is not needed, you have `translation` computed in line 19

Revit_Core_Engine/Query/TransformToOriginInXY.cs
```diff
@@ -0,0 +1,42 @@
...
+            List<Point> pts = outline?.ControlPoints;
+            if (pts == null || pts.Count < 3)
+                return (translation, rotation);
+
+            int n = pts.Count;
+            if (n >= 3 && (pts[n - 1] - pts[0]).Length() <= BH.oM.Geometry.Tolerance.Distance)
+                n--;
+
+            if (n < 3)
+                return (translation, rotation);
```
pawelbaran: These lines are not needed, the check is implicitly made in `Centroid` query

pawelbaran (commented):
Looks quite good now, please see the code comments + there are compliance issues to fix as well.

Revit_Core_Engine/Query/PadFoundation.cs
```diff
@@ -0,0 +1,141 @@
...
+            return element?.FoundationGeometryOutline()?.Centroid();
+        }
+
+        /***************************************************/
+
+        [Description("Checks whether a Polyline represents a rectangle.")]
+        [Input("polyline", "Polyline to check.")]
+        [Input("settings", "Revit adapter settings containing tolerance values. If null, default tolerance will be used.")]
+        [Output("isRectangular", "True if the polyline represents a rectangle; otherwise false.")]
+        public static bool IsRectangle(this Polyline polyline, RevitSettings settings = null)
```
pawelbaran: Why do we still have this method here if it was migrated to Geometry_Engine?

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,141 @@
...
+            double diagonal2 = pts[3].Distance(pts[1]);
+
+            return Math.Abs(diagonal1 - diagonal2) <= tol;
+        }
+
+        /***************************************************/
+        /****              Private methods              ****/
+        /***************************************************/
+
+        private static Vector LongestEdgeDirection(this Polyline polyline, double tolerance, double angleTolerance = Tolerance.Angle)
```
pawelbaran: Why do we still have this method here if it was migrated to Geometry_Engine?

Revit_Core_Engine/Query/FoundationGeometry.cs
```diff
@@ -0,0 +1,141 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Extracts the outer rectangular boundary of a PadFoundation as a Polyline.")]
+        [Input("element", "PadFoundation element whose boundary should be extracted.")]
+        [Output("outline", "Polyline representing the PadFoundation external boundary.")]
+        public static Polyline FoundationGeometryOutline(this PadFoundation element)
```
pawelbaran: I would consider renaming this file to PadFoundation.cs, then the method name prefix from `FoundationGeometry` to `PadFoundation`.

pawelbaran (changes requested):
The code stopped compiling after the last commit

Revit_Core_Engine/Modify/SetLocation.cs
```diff
@@ -344,6 +345,101 @@ public static bool SetLocation(this FamilyInstance element, Pile pile, RevitSett
...
+            Polyline bhomOutline = padFoundation.PadFoundationOutline();
+            if (bhomOutline == null)
+                return false;
+
+            // Transformation needed to bring BHoM outline to global XY
+            (Vector, double) bhomTransform = bhomOutline.TransformToOriginInXY();
+            if (bhomTransform.Item1 == null || double.IsNaN(bhomTransform.Item2))
+                return false;
+
+            // Translation between the current location and BHoM centroid
```
pawelbaran:
```suggestion
            // Translation between the origin and BHoM centroid
```

Revit_Core_Engine/Modify/SetLocation.cs
```diff
@@ -344,6 +345,101 @@ public static bool SetLocation(this FamilyInstance element, Pile pile, RevitSett
...
+            XYZ referencePoint = null;
+            foreach (Autodesk.Revit.DB.Face face in element.Faces(new Options(), settings))
+            {
+                if (face is PlanarFace planarFace && Math.Abs(1 - planarFace.FaceNormal.DotProduct(XYZ.BasisZ)) <= BH.oM.Adapters.Revit.Tolerance.Angle)
+                {
+                    referencePoint = planarFace.Centroid();
+                    break;
+                }
+            }
+
```
pawelbaran, line 387: Worth adding a null check against `referencePoint` here

Revit_Core_Engine/Modify/SetLocation.cs
```diff
@@ -344,6 +345,101 @@ public static bool SetLocation(this FamilyInstance element, Pile pile, RevitSett
...
+            {
+                Autodesk.Revit.DB.Line axis = Autodesk.Revit.DB.Line.CreateBound(bhomXY, bhomXY + XYZ.BasisZ);
+                ElementTransformUtils.RotateElement(element.Document, element.Id, axis, dRot);
+                updated = true;
+                element.Document.Regenerate();
+            }
+
+            BoundingBoxXYZ bbox = element.get_BoundingBox(null);
+            double topZ = bbox.Max.Z;
+            BH.oM.Geometry.Point bhomCentroid = padFoundation.PadFoundationCentroid();
```
pawelbaran: No need to calculate it here, can simply take `bhomTransform.Item1.Z` (or rather negative of it)

Revit_Core_Engine/Query/IsMatchingOutline.cs
```diff
@@ -0,0 +1,123 @@
...
+        /***************************************************/
+
+        [Description("Checks whether the given BHoM Polyline outline matches the outline of the padFoundationFamily freeform extrusion in Revit.")]
+        [Input("padFoundationFamily", "Revit padFoundationFamily whose extrusion outline should be compared.")]
+        [Input("orientedOutline", "BHoM polyline outline to match against the padFoundationFamily outline.")]
+        [Input("settings", "Revit adapter settings providing distance tolerance for the matching.")]
+        [Output("matches", "True if the padFoundationFamily extrusion outline matches the input outline.")]
+        public static bool IsMatchingOutline(this Family padFoundationFamily, Polyline orientedOutline, RevitSettings settings)
+        {
+            Document document = padFoundationFamily.Document;
```
pawelbaran: this variable is not needed anywhere

Revit_Core_Engine/Query/TransformToOriginInXY.cs
```diff
@@ -0,0 +1,57 @@
...
+using BH.Engine.Geometry;
+using BH.oM.Base.Attributes;
+using BH.oM.Geometry;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Query
+    {
```
pawelbaran, line 34: /* Public Methods */ decorator missing

pawelbaran (changes requested):
Starts looking very good overall!! I left a few comments in the code, also one issue with the tests:
1. Select the upper left rectangle

Image

2. Update polygon outline

Image

3. Rhino vs Revit (mirror against Y):

Image

pawelbaran:
@BHoMBot check compliance

pawelbaran (approved):
Code reviewed & tested, really happy with it now, thank you @WojciechBuda00 💪

pawelbaran:
@BHoMBot check required

PR #1675, opened 2026-07-14
Push of PileFoundation added

### LINKED ISSUE #1674: Added Push PileFoundation

WojciechBuda00:

Closes #1674 

### Test files

[#1674-PushPileFoundation](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/Revit_Toolkit/%231674-PushPileFoundation?d=wfb6fbd62b1f145fdb01f0b0acd776d82&csf=1&web=1&e=5oKnRk)


Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+                return null;
+            }
+
+            // Get the thickness and check if valid
+            double thickness = pileFoundation.PileCap.PadFoundationThickness();
+            if (double.IsNaN(thickness))
+                return null;
+
+            //Get the pielDepth and check if valid
+            double pileDepth = pileFoundation.Piles.Select(p => p.Location as BH.oM.Geometry.Line).Where(l => l != null).Select(l => Math.Abs(l.Start.Z - l.End.Z)).DefaultIfEmpty(double.NaN).Max();
```
pawelbaran: Would be good to add a check if all piles are vertical and same top/bottom Z - if not, raise an error and return null?

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        [Description("Generates a Revit FamilySymbol for a BHoM PileFoundation by querying existing freeform families in the document or creating a new one from a template file.")]
+        [Input("pileFoundation", "BHoM pile foundation to generate the Revit type for.")]
+        [Input("document", "Revit document, in which the family type will be created.")]
+        [Input("settings", "Settings to be used when generating the family type.")]
+        [Output("symbol", "Created Revit family type that represents the input BHoM pile foundation.")]
+        public static FamilySymbol GeneratePileFoundationType(this PileFoundation pileFoundation, Document document, RevitSettings settings = null)
```
pawelbaran, line 55: Would be good to check if there are any piles and return pad foundation type in case if not

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+            if (double.IsNaN(thickness))
+                return null;
+
+            //Get the pielDepth and check if valid
+            double pileDepth = pileFoundation.Piles.Select(p => p.Location as BH.oM.Geometry.Line).Where(l => l != null).Select(l => Math.Abs(l.Start.Z - l.End.Z)).DefaultIfEmpty(double.NaN).Max();
+            if (double.IsNaN(pileDepth))
+                return null;
+
+            //Get the diameter of first nested pile
+            CircleProfile profile = (pileFoundation.Piles[0].Property as ConstantFramingProperty)?.Profile as CircleProfile;
```
pawelbaran:
```suggestion
            CircleProfile profile = (pileFoundation.Piles.Select(x => x.Property).OfType<ConstantFramingProperty>().Select(x=>x.Profile).OfType<CircleProfile>();
```

I'd rather do as follows:
1. pull out all circle profiles (see above)
2. check if the number is equal to number of piles, if not then error & return null
3. check if all diameters are the same, if not then error & return null

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+                List<int> takenIndices = freeformFamilies.Select(x => Regex.Match(x.Name, $"{Regex.Escape(prefix)}(\\d+)$")).Select(x => int.Parse(x.Groups[1].Value)).ToList();
+                int newIndex = takenIndices.Count > 0 ? takenIndices.Max() + 1 : 1;
+                family = GenerateFreeFormPileFoundationFamilyFromTemplate(document, orientedOutline, layout, diameter, pileDepth, newIndex, pileFoundation, settings);
+            }
+
+            if (family == null)
+                return null;
+
+            // Find or create the type with matching thickness, diameter and pileDepth
+            return family.FindOrCreatePileTypeWithDimensions(thickness, diameter, pileDepth);
```
pawelbaran: No need for diameter here any more? As agreed, it should be hardcoded inside the family?

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        /***************************************************/
+
+        private static Family GenerateFreeFormPileFoundationFamilyFromTemplate(this Document document, Polyline orientedOutline, ExplicitLayout layout, double diameter, double pileDepth, int index, PileFoundation pileFoundation, RevitSettings settings = null)
```
pawelbaran: it feels like the input of type `PileFoundation` is not needed here - also please recall our conversation about the sister method in `PadFoundation`, there we also figured that the input of type `PadFoundation` was not needed.

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        /***************************************************/
+
+        private static FamilySymbol PrepareNestedPileSymbol(FamilySymbol pileSymbol, double diameter, double pileDepth)
+        {
+            if (pileSymbol == null || diameter <= 0)
+                return pileSymbol;
+
+            string typeName = $"{(long)Math.Round(diameter * 1000.0)}Ø";
```
pawelbaran: Isn't there a chance to reuse something from the pile convert? e.g. `Pile.ElementType(Document)` extension method? Feels a bit like reinventing the wheel (but not saying it is not necessary)

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
```diff
@@ -0,0 +1,369 @@
...
+
+            if (!nestSymbol.IsActive)
+                nestSymbol.Activate();
+
+            return nestSymbol;
+        }
+
+        /***************************************************/
+
+        private static FamilySymbol FindOrCreatePileTypeWithDimensions(this Family family, double thickness, double diameter, double pileDepth)
```
pawelbaran: This is not pile type, rather pile foundation type. i'd consider renaming as below


```suggestion
        private static FamilySymbol FindOrCreateTypeWithDimensions(this Family family, double thickness, double diameter, double pileDepth)
```

Revit_Core_Engine/Query/PileFoundation.cs
```diff
@@ -0,0 +1,61 @@
...
+using BH.oM.Geometry;
+using BH.oM.Physical.Elements;
+using BH.oM.Spatial.Layouts;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace BH.Revit.Engine.Core
+{
+    public static partial class Query
+    {
```
pawelbaran, line 38: Missing `/*******` decorators

pawelbaran (commented):
Initial code review so far, will do the testing now and continue digging in the code 👍

pawelbaran:
@BHoMBot check compliance
@BHoMBot check core

Revit_Core_Engine/Convert/Physical/ToRevit/FamilyInstance.cs
```diff
@@ -273,6 +273,56 @@ public static FamilyInstance ToRevitFamilyInstance(this PadFoundation padFoundat
...
+
+            settings = settings.DefaultIfNull();
+
+            if (pileFoundation.PileCap == null)
+            {
+                BH.Engine.Base.Compute.RecordError($"PileFoundation has no PileCap. BHoM_Guid: {pileFoundation.BHoM_Guid}");
+                return null;
+            }
+
+            FamilySymbol symbol = pileFoundation.GeneratePileFoundationType(document, settings);
```
pawelbaran: This should be `ElementType` query that would call `GeneratePileFoundationType`, similar to `PadFoundation`

pawelbaran (changes requested):
More general comments:
- you need to merge `develop` into your branch, now yours is far behind `develop`
- from what I saw in the source family, pile depth should be instance parameter, not type? if that's the case, you will need to set pile depth explicitly, both in create and update
    Image
- pile foundation pushed from GH looks as follows:
    Image
    ...after pulling back from Revit, it is like this:
    Image
    ...the piles are much longer, which is caused by the fact that you are calculating pile depth as pile top minus pile bottom, which is not correct: the piles in GH are a simplification snapped to the upper face of the cap. Instead, I would try the following:
    1. find top and bottom of the pile
    2. check if top elevation == bottom of the pile cap => if not, warning that the pile depth has been set based on bottom of the cap and bottom elevation of the pile
    3. set the depth = bottom of the pile cap - pile bottom
- I would consider tweaking the naming:
    - pile foundation, I would consider renaming the type to `{thickness}mm thk. Ø{diameter}`, assuming that, as above, pile depth is instance param
        Image
    - nested pile, freeform suffix looks suspicious
        Image
- provided test script does not show any update, also top foundation (square) produces 2 foundations with 2 piles each instead of 1 foundation with 4 piles (missing flatten)

Revit_Core_Engine/Query/IsMatchingOutlineAndLayout.cs
```diff
@@ -0,0 +1,143 @@
...
+            return false;
+        }
+
+        /***************************************************/
+
+        private static List<oM.Geometry.Point> PileLayoutPoints(this Family family, RevitSettings settings)
+        {
+            Document doc = family.Document;
+
+            FamilyInstance host = new FilteredElementCollector(doc)
```
pawelbaran, line 117: Also this will only work when the instance is already placed in the model. Alternatively, you could get the first `FamilySymbol` in the `Family` and get the piles out via `GetDependentElements` - would that work?

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
WojciechBuda00: No change for now, looks necessary to me

Revit_Core_Engine/Compute/GeneratePileFoundationType.cs
WojciechBuda00: Following our conversation, we keep diameter for naming.

Revit_Core_Engine/Query/IsMatchingOutlineAndLayout.cs
WojciechBuda00: As discussed, this will be developed in the next update.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core
@BHoMBot check null-handling
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check required

pawelbaran (approved):
Code reviewed and tested, happy to approve 👍

pawelbaran:

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: ready-to-merge

pawelbaran:
@sakanni FYI I've just set the rtm check to neutral as per offline chat

PR #1681, opened 2026-07-30
Updated Connector Query to fallback to max distance if connector has no Origin, eg Logical Connectors

### LINKED ISSUE #1680: MEP Connectors Query method fails for logical connectors
#### Description:
When a system path includes logical connectors, any methods that rely on sorting connectors will throw and exception when .Origin is called.

#### Expected behaviour:
Include Logical connectors as last in the sorting order, OR exclude them from the returned sorted connectors. I favor the first option.

enarhi:

Closes #1680 

The method threw an exception if it was called with Logical Connectors. This update places Logical Connectors at the end of the sorted list instead.

### Test files
Any Revit File with MEP Systems containing Logical Connectors.

### DIFF
--- a/Revit_Core_Engine/Query/Connectors.cs
+++ b/Revit_Core_Engine/Query/Connectors.cs
@@ -58,7 +58,12 @@ namespace BH.Revit.Engine.Core
             foreach (Connector conn in connSet)
                 connList.Add(conn);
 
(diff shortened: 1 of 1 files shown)


michal-pekacki (approved):
Happy with the fix for connectors where `Origin` isn't available 👍

michal-pekacki:
@BHoMBot check ready-to-merge

PR #1685, opened 2026-08-03
Level property not found properly

### LINKED ISSUE #1684: Level property not found properly
#### Description:

When there are no parameters that are not read only, no value is picked up and output is set to null.

adam-sobieski:
Level.cs: introduce a local Level variable and change lookup flow to prefer LevelParameter and family host, then fall back to Document.LevelBelow(location) instead of returning null. LevelParameter.cs: adjust LINQ selection to use ThenByDescending on IsReadOnly and call FirstOrDefault() (remove predicate) so the candidate is chosen via ordered ranking (valid ElementIds first, then read-only ordering). This makes level resolution more robust and centralizes parameter selection logic.

Closes #1684

### DIFF
--- a/Revit_Core_Engine/Query/Level.cs
+++ b/Revit_Core_Engine/Query/Level.cs
@@ -42,31 +42,32 @@ namespace BH.Revit.Engine.Core
             if (element == null)
                 return null;
 
+            Level level = null;
             Document doc = element.Document;
             ElementId levelId = element.LevelId;
 
             // First priority: Check LevelId property (most direct way to get level)
             if (levelId.Value() != -1)
-                return doc.GetElement(levelId) as Level;
-
-            Level level = null;
+                level = doc.GetElement(levelId) as Level;
+            if (level != null)
+                return level;
 
             // Second priority: Check level parameters (LevelParameter or BaseLevelParameter)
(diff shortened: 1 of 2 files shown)


Revit_Core_Engine/Query/Level.cs
```diff
@@ -49,17 +49,25 @@ public static Level Level(this Element element)
...
+            if (level != null)
+                return level;
 
             // Third priority: Check if element is work plane-based and hosted on a level
             if (element.IsWorkPlaneLevelBased())
-                return (element as FamilyInstance)?.Host as Level;
+                level = (element as FamilyInstance)?.Host as Level;
+
+            if (level == null)
+                level = element.Document.LevelBelow(element.LocationPoint());
```
michal-pekacki:
```suggestion
 // Fourth priority: Check closest level below
            if (level == null)
                level = element.Document.LevelBelow(element.LocationPoint());
```

michal-pekacki (approved):
Changes make sense, happy to approve 👍

adam-sobieski:
@BHoMBot check required

michal-pekacki:
@BHoMBot check versioning

michal-pekacki:
@BHoMBot check ready-to-merge

PR #1686, opened 2026-08-14
ci: point the CI workflow at BHoM/CI_Toolkit

sakanni:
CI_Toolkit moved from BuroHappoldEngineeringAdmin to BHoM. The old path redirects, so this removes a dependency on the redirect rather than fixing a break.

Also picks up two pattern corrections made since this file was copied: dataset compliance now uses a case-insensitive unanchored pathspec, and project compliance matches `AssemblyInfo.cs` at any depth instead of only at the repo root. Neither changes which files are selected in this repo today.

Still non-blocking. Tracked in CI_Toolkit_Proxy#20.

### DIFF
--- a/.github/workflows/ci-beta.yml
+++ b/.github/workflows/ci-beta.yml
@@ -21,7 +21,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Run build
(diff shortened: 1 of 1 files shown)


peterjamesnugent (approved):
Happy with changes just repointing the path to BHoM/CI_Toolkit as evidenced by the checks passing,

sakanni:
@BHoMBot check ready-to-merge

